PR #12, opened 2016-06-28
Nalipiev/grunt test config

Lipata:
Add test and coverage configuration. Change test references. Please create new issue to fix all the test.


damyanpetev:
@kdinev Can you add this repo in [coveralls](https://coveralls.io/github/igniteui) and #5 should be done as the module did try to submit results and got the missing repo response.
The build is currently running only editors tests (as some need to be fixed separately) and there is a temp force due to coverage warnings for running only those.

kdinev:
@damyanpetev done.

PR #19, opened 2016-06-29
Nalipiev/grunt test config

Lipata:


kdinev:
Apparently, references to the `igDataSource` are wrong in some of the test files.

PR #24, opened 2016-06-30
Cannot read property dpDiv of undefined is thrown after setting MinDa…

dafo:
#27 
 Exception when setting min/max date


kdinev:
Reference a publicly visible issue. If the issue exists only in our internal system, then mirror it on GitHub.

dafo:
#27
Exception when setting min/max date

PR #25, opened 2016-06-30
Fixing issue with igDatePicker - open dropdown if readOnly and dropDo…

Marina-L-Stoyanova:
…wnOnReadOnly are set to true.


kdinev:
@Marina-L-Stoyanova Can you create a public issue about this and reference it in the pull request?

Marina-L-Stoyanova:
Closes #28

PR #32, opened 2016-06-30
Zoombar custom provider for zoomable components

### LINKED ISSUE #15: Make the chart provider for the igZoombar configurable
Currently the default charting provider for the igZoombar is `$.ig.ZoombarProviderDataChart` but there is no straightforward way to swap this with a different provider. Make the provider be passed into the zoombar as an option with the same default value.

`chartProvider: $.ig.ZoombarProviderDataChart`

### LINKED ISSUE #21: Fix failing unit tests
- [x] igDataSource
- [x] igCombo
- [x] igZoombar
- [ ] Formatter
- [x] Editors

ChronosSF:
Closes #15 Closes #21 Changing the option type to provider to facilitate the user providing
custom interfaces for different chart components. Fixing tests.


kdinev:
@ChronosSF this is a breaking change because you're changing an option name. Either keep the option name to `type`, or add documentation about this breaking change. If you're going to remove the `type` option, then you have to deprecate it. You cannot just remove it.

ChronosSF:
It's a breaking change and not. The option served little purpose as all its valid values would simply lead to a DataChart provider being initialized (and invalid would throw an exception). Technically even if users would set it the widget will simply ignore it at the moment and render a DataChart anyway - so no behavioral change would occur.

(the option was a placeholder so that igZoombar had the extensibility built-in for us to add more providers in the future - it was basically a provider switch, however, it was not capable to support custom providers - we need a better name for such an option)

The option is not mentioned in the docs anywhere (except API docs of course but they'll be updated with the build)

kdinev:
@ChronosSF That makes sense. Merging.

kdinev:
@ChronosSF tests are timeout-ing.

PR #42, opened 2016-07-01
Fixes on QUnit procedure.

Lipata:


kdinev:
This change was done directly in master.

PR #64, opened 2016-07-05
Prevent commit failure from npm/bower version calls

damyanpetev:
before changes are finalized,  #7


damyanpetev:
@kdinev can you tag again after merging this?

Issue #68, opened 2016-07-06
igZoombar has jQuery UI-specific code when communicating with the target chart


ChronosSF:
A few more changes are needed for the Zoombar to be truly agnostic towards the component that's being zoomed.

PR #82, opened 2016-07-07
Zoombar custom chart provider improvements

ChronosSF:
Changes for issue #68 - cleaning the Zoombar from mode provider-specific code, adding callbacks (instead of directly binding to chart events), removing unnecessary provider settings.


kdinev:
The API methods are missing parameter types and descriptions. Some of them are not even documented.

PR #97, opened 2016-07-13
Fix _processGroupsRecursive to properly generate id of the grouped re…

m-hristov:
…cord


kdinev:
@m-hristov If IDs were not properly generated previously, and there were no failing unit tests, then that means a unit test needs to be written for this scenario!

Issue #105, opened 2016-07-14
Increase colorpicker and colorpickersplitbutton code coverage


zdrawku:
Coverage should go up to 90%+

- [x] Colorpicker
- [x] Colorpicker split button

PR #111, opened 2016-07-18
issue #105 colorpicker and colorpickersplitbutton code coverage increased

Aleksandyr:


kdinev:
@Aleksandyr The error message parameter for asserts is what error should be displayed. 

Example:

`equal(this.actualColorPicker.options.colors.length , 0, "array of colors set to empty");` 

In this case the assert is that the colors are correctly removed, meaning set to an empty array. The error message is displayed if this assert does not pass, meaning there are still colors in the color picker. In this case the error message should rather read `"Array of colors is not set to empty via setOption."`

`equal(this.actualColorPicker.options.colors.length , 0, "Array of colors is not set to empty via setOption.");` 

In addition the `_setOption` rarely ever fails, because it's implemented in the base widget. There should be additional asserts checking that the DOM elements were removed, etc.

tests/unit/colorpicker/tests.html
```diff
@@ -48,44 +48,49 @@
 
 	    test("_changeSelectedColor", function () {
 	        this.actualColorPicker._changeSelectedColor(this.firstColor);
-	        ok(this.firstColor.hasClass("selected-color"), "first color selected");
+	        ok(this.firstColor.hasClass("selected-color"), "first color wasn't selected");
 	    });
 
 	    test("_changeColorOptions", function() {
-	    	this.actualColorPicker.element.igColorPicker("option", "colors", []);
```
kdinev: This calls _setOption. Don't change it!

tests/unit/colorpicker/tests.html
```diff
@@ -48,44 +48,49 @@
...
-	    	equal(this.actualColorPicker.options.colors.length , 0, "array of colors set to empty");
+	    	this.actualColorPicker._setOption("colors", []);
+	    	equal(this.actualColorPicker.options.colors.length , 0, "colors options is set to empty through _setOption method");
 	    })
+	    
 	    test("_changeStandardColorsOption", function () {
 	    	var colors = ["#f0ff0f", "#000a00", "#FEECE1"];
 	    	
-	    	notEqual(this.actualColorPicker.element.igColorPicker("option", "standardColors"), colors, "arry of given colors isn't equal to standard colors");
+	    	notEqual(this.actualColorPicker.options.standardColors, colors, "arry of given colors isn't equal to  standardColors option");
```
kdinev: This calls _getOption, don't change it!

kdinev:
I notice some general issues with the control, that are reflected in the tests. There are too many things that needs to be available as public API, but are missing and thus are accessed in the test as private calls. Rework the control code and add the missing public API.

tests/unit/colorpicker/tests.html
```diff
@@ -43,54 +45,61 @@
...
 	    	
-	    	notEqual(this.actualColorPicker.options.standardColors, colors, "standardColors options is equal with givan colors");
+	    	notEqual(this.actualColorPicker.igColorPicker("option", "standardColors"), colors, "standardColors options is equal with givan colors.");
 	    	
-	    	this.actualColorPicker._setOption("standardColors", colors)
-	    	equal(this.actualColorPicker.options.standardColors, colors, "array of standardColors option was not changed through _setOption method");
+	    	this.actualColorPicker.igColorPicker("option", "standardColors", colors)
+	    	deepEqual(this.actualColorPicker.igColorPicker("option", "standardColors"), colors, "array of standardColors option was not changed."); // deepEqual
+
+	    	equal(this.actualColorPicker.igColorPicker("option", "standardColors").length, colors.length, "standard colors length isn't equal with given colors length.");
```
kdinev: Here you should include the same test as above, where you retrieve the standardColorsTable and check the length of the children. This will require one more API method, I guess :)

tests/unit/colorpicker/tests.html
```diff
@@ -43,54 +45,61 @@
...
 	    	
-	    	this.actualColorPicker._setOption("standardColors", color)
-	    	equal(this.actualColorPicker.options.standardColors, color, "given color was not set to standardColors through _setOption method");
+	    	this.actualColorPicker.igColorPicker("option", "standardColors", color)
+	    	equal(this.actualColorPicker.igColorPicker("option", "standardColors"), color, "given color was not set to standardColors.");
 	    	
-	    	this.actualColorPicker._setOption("standardColors", color)
-	    	equal(color, this.actualColorPicker.options.standardColors, "standardColors option value has different value from given color value");
+	    	this.actualColorPicker.igColorPicker("option", "standardColors", color)
+	    	equal(color, this.actualColorPicker.igColorPicker("option", "standardColors"), "standard color value isn't equal with given color value.");
```
kdinev: Same here. Check the DOM elements and use the new API method for this.

tests/unit/colorpicker/tests.html
kdinev: Also check that clicking on the new colors selects them!

Issue #114, opened 2016-07-18
Validator support for Credit Card check


damyanpetev:
Partially inspired by Data Annotation in MVC, but also seemed reasonable to support such common check.
**New option:**
`creditCard` - bool or object with `errorMessage`, similar to email and other type checks. Inherited.
**Description:**
Validation is client side and credit card type agnostic - as [all of the active issuers](https://en.wikipedia.org/wiki/Payment_card_number) can be validated using the Luhn algorithm. Check will match the MVC server side model validation and will be accompanied by a respective option in the Validator Model/Wrapper.

PR #118, opened 2016-07-18
Validator creditcard

damyanpetev:
- Adding `creditCard` option and validation rule for #114 
  - CreditCard comes before pattern/custom rules in case it has to validate before more fine-grained checks (per card type) are performed in those. Inherited.
 -  Using "payment card number" for user-facing error message as it is the official general term even though credit card might be more commonly used, but keeping the latter as option name to match expectations from popular frameworks.
- Extending string with reverse() function in util (searched again, don't see another)

I did a bit of refactoring on the validation rules to reduce the bulk that was the `_validateField` method. Also makes it easy to plug in new rules or override current ones. Tried to decouple them completely from the validator, but that ended up being too much hassle since the custom rule needs to call function in the validator context and the EqualsTo rule needs the value from target function. Also allowed rules to keep some state between validations to be able to format messages without explicitly using the validator functions to get localized message.

Most of the logic for rules remains unchanged, with two notable changes:
- Number checking regex is gone as it was too limited, replaced by new NaN checks and those are now inherited by the Value Range rule as well.
- Field options passed to "custom" rule option now contain inherited options.

Questions:
- Since we support defining decimal/thousands separator in `number` option, I've reused the parsing if such option is defined when checking value range - it only seemed fitting if you expect numbers in a certain format to expect range checks to be able to work with those too. Would it make sense to just use that parsing by default regardless?
- Above separators as well as a regex for email are now part of their rules, but there's also a `$.ui.igValidator.defaults` object (legacy) that could accommodate all such settings. Yay or nay?
- There's  a backup locale object in the validator prototype in case there are no options and locale file included. 	I've kept it as a legacy thing in the 15.2 rewrite, but those string are either duplicate or outdated and would actually like to see them removed.

 @bazal4o, @mpavlinov and @kdinev feel free to review.


kdinev:
@damyanpetev 
1. Yes
2. Yay
3. Remove 👍 

Also, after the PR is merged, you should assign the issue to localization for translation, and notify them.

damyanpetev:
@kdinev All finishing touches are done.
Also decided to allow the card rule to inherit, for consistency with other type-like rules.

Issue #101, opened 2016-07-14
igSplitter makes the second panel take the whole width if width for the first panel is not set


hanastasov:
If we initialize igSplitter as follows:

```
$("#splitter").igSplitter({
    height: 700,
    panels: [{}, { "size": 250}]
});
```

then the first panel is not visible and second panel takes the whole available width.

PR #134, opened 2016-07-21
Merge; #101

petko5kov:
Fix for #101


kdinev:
@petko5kov You should add a unit test for the issue you're fixing ;)

Issue #52, opened 2016-07-04
Increase combo code coverage


rkaraivanov:
Coverage should go up to 90%+

PR #136, opened 2016-07-22
increase combo code coverage

hanastasov:
#52


hanastasov:
Obviously the last one will not make it for the IgniteUI OSS :)). I missed to consider this in advance.

kdinev:
Yeah, that should be removed.

Issue #103, opened 2016-07-14
Increase htmlEditor code coverage


zdrawku:
Coverage should go up to 90%+

PR #144, opened 2016-07-26
Increase igHtmlEditor coverage over 75%

zdrawku:
#103

### DIFF
--- a/tests/unit/htmleditor/basicfunctionalities/rendering/tests.html
+++ b/tests/unit/htmleditor/basicfunctionalities/rendering/tests.html
@@ -325,7 +325,7 @@
                 }, 700);
             },
             adjustDomPathToolbarWidth: function(){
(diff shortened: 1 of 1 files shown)


tests/unit/htmleditor/basicfunctionalities/rendering/tests.html
```diff
@@ -325,7 +325,7 @@
                 }, 700);
             },
             adjustDomPathToolbarWidth: function(){
-                var htmlEditor1 = $('#' + this.adjustSourceButtonsEditorId), contentToSelect,
+                var self = this, htmlEditor1 = $('#' + this.adjustSourceButtonsEditorId), contentToSelect,
```
mpavlinov: I don't see `self` used in a closure.

zdrawku: I've removed it. Thanks

PR #155, opened 2016-07-29
Adding additional tests for different data source types

MayaKirova:
Adding additional tests for different data source types

### DIFF
--- a/tests/unit/dataSource/properties/tests.html
+++ b/tests/unit/dataSource/properties/tests.html
@@ -126,6 +126,62 @@
 				$.ig.DataSourceLocale.locale.errorParsingJsonNoSchema + " should be thrown"
 			);
 		});
+		
+		test("Test creating JSONDataSource", function () {
+			 var adventureWorks = {
+                "Records": [{
+                    "ProductID": 1,
+                    "Name": "Adjustable Race",
+                    "ProductNumber": "AR-5381"
+                }, {
+                    "ProductID": 2,
+                    "Name": "Bearing Ball",
+                    "ProductNumber": "BA-8327"
+                }
+            ]};
+			 var jsonpDs = new $.ig.JSONDataSource({
+                        dataSource: adventureWorks,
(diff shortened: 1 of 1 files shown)


tests/unit/dataSource/properties/tests.html
```diff
@@ -126,6 +126,62 @@
...
+			equal(jsonpDs.type(), "json", "Check if dataSource type is correct.");
+			equal(jsonpDs.dataView().length, 2, "DataView should contain 2 records.");			
+			
+			jsonpDs = new $.ig.JSONDataSource();
+
+			equal(jsonpDs.type(), "json", "Check if dataSource type is correct.");			
+		});
+		
+		test("Test creating HtmlTableDataSource ", function () {
+			var htmltableDs = new $.ig.HtmlTableDataSource ();
```
mpavlinov: Feed it with some HTML and see if it gets parsed correctly.

tests/unit/dataSource/properties/tests.html
```diff
@@ -126,6 +126,62 @@
...
+			equal(htmltableDs.type(), "htmlTableDom", "Check if dataSource type is correct.");			
+		});
+		
+		test("Test creating ArrayDataSource  ", function () {
+			var arrayDs = new $.ig.ArrayDataSource();
```
mpavlinov: Feed it with an array data and see if it is available in the dataView()

tests/unit/dataSource/properties/tests.html
```diff
@@ -126,6 +126,62 @@
...
+			equal(arrayDs.type(), "array", "Check if dataSource type is correct.");			
+		});
+		
+		test("Test creating RemoteDataSource   ", function () {
+			var arrayDs = new $.ig.RemoteDataSource ();
```
mpavlinov: Same here. Give it an Url.

MayaKirova:
There are already similar tests under dataSource/schemas so there's no need to verify this again.

Issue #63, opened 2016-07-05
Increase datasource code coverage


kdinev:
Coverage should go up to 90%+

PR #168, opened 2016-08-02
Adding more tests for igDataSource

MayaKirova:
Adding more tests for igDataSource
#63


tests/unit/dataSource/properties/tests.html
```diff
@@ -453,7 +532,20 @@
 			equal(dsLocal.pendingTransactions().length, 0, "There should be a no pending transaction after the transaction has been rolled back.");
 		});
 
+		test("Test _removeTransactionsByRecordId API", function () {
+			var transaction = dsLocal.addRow(999, { "ID": 999, "Name": "Bread", "Price": 2.5 }, false);
```
mpavlinov: You should have more than one transaction pending, so that you can see whether the correct transaction is removed.

tests/unit/dataSource/properties/tests.html
```diff
@@ -536,22 +628,113 @@
...
-			equal(rec.Name, "NewVal", "Detached value should be updated.");
+			dsLocal.dataBind();
+			stop();
+
+			setTimeout(function () {
+				start();
+				var t = dsLocal.setCellValue(0, "Name", "NewVal", true);
+				var rec = dsLocal.getDetachedRecord(t);
+
+				equal(rec.Name, "NewVal", "Detached value should be updated.");
```
mpavlinov: You should probably call setCellValue once again on the same record and then check that the detached record is not updated.

PR #198, opened 2016-08-05
Chart loader update

alexkartavov:
The changes include new modules for the DV widgets.
It also optimizes loader to use $.holdReady to defer firing jQuery's document ready event until Ignite's resources are loaded.


src/js/infragistics.loader.js
```diff
@@ -135,135 +212,188 @@ $.ig.dependencies = [
...
 		group: $.ig.loaderClass.locale.dvGroup,
-		scripts: [ "$path$/modules/infragistics.chart_annotationchart.js" ]
+		scripts: [ "$path$/modules/infragistics.datachart_annotation.js" ]
 	},
 	{
 		widget: "igDataChart.*",
 		dependency: [ { name: "Category" },
+					{ name: "RangeCategory" },
+					{ name: "VerticalCategory" },
 					{ name: "Financial" },
```
mpavlinov: "ExtendedFinancial" dependency is missing.

PR #201, opened 2016-08-08
Fix issue: Child data persists in the datasource after deleting the corresponding parent record in treegrid 

m-hristov:


mpavlinov:
Please, add some unit tests.

PR #204, opened 2016-08-08
Fix issue #203 - Toolbar button does not work properly for a selected content range if a text is initially selected by double-tapping, and the selection is changed by mouse dragging.

### LINKED ISSUE #203: Toolbar button does not work properly for a selected content range if a text is initially selected by double-tapping, and the selection is changed by mouse dragging.
Toolbar button does not work properly for a selected content range if a text is initially selected by double-tapping, and the selection is changed by mouse dragging.

Aleksandyr:
This fixes #203

### DIFF
--- a/src/js/modules/infragistics.ui.htmleditor.js
+++ b/src/js/modules/infragistics.ui.htmleditor.js
@@ -2735,10 +2735,13 @@ if (typeof jQuery !== "function") {
                     }
                 }
 
+                // A.K August 8th, 2016 Bug #219768 Toolbar button does not work properly for a selected content range if a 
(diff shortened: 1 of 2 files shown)


mpavlinov:
There are some jsHint errors that you need to fix.

mpavlinov:
Now there are some JSCS errors that you need to fix.

Aleksandyr:
Should i fix the errors in infragistics.util.js file, the changes there isn't from my fix, but it fails again?

mpavlinov:
Only the ones in the igHtmlEditor

mpavlinov:
Now you need to add some unit test on the code that you added.

PR #206, opened 2016-08-09
Chart loader update

alexkartavov:
Some minor fixes to the loader needed for the script combiner. Thanks!


mpavlinov:
There is a JSCS error.

alexkartavov:
Thanks. This is fixed now

Issue #230, opened 2016-08-17
[Util] $.ig.getColType function returns incorrect result for Date object 


MayaKirova:
[Util] $.ig.getColType function returns incorrect result for Date object

PR #232, opened 2016-08-17
Fixing issue in $.ig.getColType function

MayaKirova:
Fixing issue in $.ig.getColType function
#230


kdinev:
@mpavlinov This should not be merged before a unit test is added!

mpavlinov:
@kdinev Ok. @MayaKirova please add unit tests for the $.ig.getColType function #230.

PR #239, opened 2016-08-18
Fixing Bug :igPercentEditor does not persist empty string value when …

### LINKED ISSUE #238: [igEditors] igPercentEditor does not persist empty string value when  allowNullValue: true
igPercentEditor does not persist empty string value when  allowNullValue: true

IvayloG:
Fixes #238


kdinev:
@IvayloG There seems to be a regression in the KO extension with this change.

PR #265, opened 2016-08-24
Fix mask validation with inner case flags and literals

### LINKED ISSUE #264: [igEditors] Mask with case flag in the middle not validated properly
1. Open [mask-w-upper-validate.zip](https://github.com/IgniteUI/ignite-ui/files/435075/mask-w-upper-validate.zip)
2. Enter "555a" in the first editor and blur
3. Do the same with the second editor

**Result**: Second editor rejects the value
**Expected**: Both editors should accept the value that covers their required field, as their masks differ only by the last two fields being either "C" or "a" flags that are both optional.

The root cause is the preemptive _validateValueAgainstMask which will not correctly move the mask position over non-char flags like ">" and "<" before checking each value char. Only reproducible when there are literal characters after said flag.

damyanpetev:
Redoing value validation against mask, moving literals and required positions before options set so validation for initial value can be handled. Tests.
fixes #264

/ cc @bazal4o

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4906,6 +4906,7 @@ if (typeof jQuery !== "function") {
 			this._promptCharsIndices = [];
 		},
 		_applyOptions: function () { //igMaskEditor
+			this._getMaskLiteralsAndRequiredPositions();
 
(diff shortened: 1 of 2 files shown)


bazal4o:
Looks good to me

PR #273, opened 2016-08-25
Fixing issues in dataSource sorting and filtering

### LINKED ISSUE #271: DataSource sorting expression with direction set to "descending" does not sort properly.
If sorting expression's dir property is set to "descending" as opposed to "desc" the data is not sorted properly.

### LINKED ISSUE #272: igDataSource filter method has a non documented parameter- fieldExpressionsOnStringsType
Parameter has no description.

MayaKirova:
Fixing issue in sortng with dir = "descending" and issue in filtering
with fieldExpressionsOnStrings parameter and documenting its purpose

closes #271
closes #272


tests/unit/dataSource/filtering/tests.html
```diff
@@ -979,11 +979,11 @@
 				});
 				test("Test function filter with argument fieldExpressionsOnStrings", function () {
 					var fd;
-					ds.filter([], null, null, "ID >= 1 AND ID < 3");
```
kdinev: @MayaKirova Shouldn't the AND logic be preserved and a new test for the OR logic be written?

MayaKirova: The old tests is not valid and needed to be changed since the purpose of that last parameter is to treat the filtered field as a string field and filter it with string conditions. 
The conditions >= and < can only apply to numeric fields. 

Since the parameter was never documented, its purpose was unclear and some incorrect changes were made to allow such a setup, however some of our other controls(like the listView) depend on that last parameter to allow filtering on fields that are not necessarily string as if they were string.

PR #282, opened 2016-08-26
Fix issue "Remote groupBy does not work in HierarchicalGrid"

m-hristov:
When grouping a column from any child grid(load on demand is NOT enabled) expanding a row to check the child grid causes a JavaScript error


kdinev:
@m-hristov is it possible to add a unit test about this?

kdinev:
@m-hristov So is it possible, or is this specific to the igGrid GroupBy feature?

m-hristov:
It is specific for group-by functionallity BUT I will try to add datasource test(later ...)

PR #291, opened 2016-08-31
Editors insert update

### LINKED ISSUE #287: [igEditors] Percent/Currency insert method not working with existing value outside edit mode
1. Open [insert-method-percent-currency.zip](https://github.com/IgniteUI/ignite-ui/files/444759/insert-method-percent-currency.zip)
2. Focus the percent editor and type 55 and focus out (insert adds 2 at the end)
3. Press the "Insert in percent" button
4. Repeat the same for currency editor

**Result:** Insert method did not change the value outside edit mode (step 3)
**Expected:** The method should change the value is some way - since there's no actual cursor when the input is not focused, the method should probably set the value instead (similar to how it worked in the editors of old) and that _should be documented in the method_.

Note: As there are two decimal positions by default, adding 2 at the end of `55.00` becomes `55.002` and gets trimmed. This is largely due to attempting to add to the display text which is why I suggest handling inserting outside edit mode as value set.

damyanpetev:
Fixes #287 
Updated common logic and method comment.

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -2973,12 +2973,12 @@
 		insert: function (string) {
 			/* Paste text at location of caret or over the current selection. Best used during editing, as the method will instead set the text as value if the editor is not focused.
 				Note: method raises the "textChanged" event.
-				paramType="string" optional="false" The string to be inserted. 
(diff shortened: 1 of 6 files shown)


damyanpetev:
Also tagging @ig-ealbert @yoshimis as the last change is to the localizable method comment as well.
/cc @bazal4o

PR #300, opened 2016-09-09
Do not set properties if the control didn't init igScroll

dkamburov:
This is related to the internal issue 224531


src/js/modules/infragistics.ui.scroll.js
```diff
@@ -3546,7 +3546,10 @@
 			container = args.owner.container().find("[data-scroll]").eq(0);
 		}
 		container.igScroll({ modifyDOM: false });
```
kdinev: But `igScroll` is initialized on the line above.

dkamburov:
It turns out there are cases when the scroll container is empty(when there is no width or height in a grid), I'll improve my commit.

Issue #295, opened 2016-09-08
Transferring code snippets 


Marina-L-Stoyanova:
- [x] igCheckboxEditor
- [x] igCurrencyEditor
- [x] igDateEditor
- [x] igDatePicker
- [x] igMaskEditor
- [x] igNumericEditor
- [x] igPercentEditor
- [x] igTextEditor
- [x] igBaseEditor
- [x] igTree
- [x] igNotifier
- [x] igValidator
- [x] igVideoPlayer
- [x] igRating
- [x] igDialog

Marina-L-Stoyanova:
As this task is done, I will close the issue.

PR #302, opened 2016-09-09
Transfer code snippets for methods and options in baseEditor

bazal4o:
#295

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -37,7 +37,7 @@
 	/* The igBaseEditor is a widget based on jQuery UI. */
 	$.widget("ui.igBaseEditor", {
 		options: {
-			/* type="string|number|null" Gets/Sets how the width of the control can be set.
+			/* type="string|number|null" Gets/Sets the width of the control.
 				```
 				//Initialize
 				$(".selector").%%WidgetName%%({
@@ -55,7 +55,7 @@
 				null type="object" will stretch to fit data, if no other widths are defined.
 			*/
 			width: null,
-			/* type="string|number|null" Gets/Sets how the height of the control can be set.
+			/* type="string|number|null" Gets/Sets the height of the control.
 				```
 				//Initialize
 				$(".selector").%%WidgetName%%({
@@ -89,7 +89,7 @@
 			```
 			*/
 			value: null,
-			/* type="number" Gets/Sets value in tabIndex for editor.
+			/* type="number" Gets/Sets tabIndex attribute for the editor input.
 			  ```
 			  //Initialize
 			  $('.selector').%%WidgetName%%({
@@ -104,7 +104,7 @@
 			  ```
 			*/
 			tabIndex: null,
-			/* type="bool" Gets/Sets ability to prevent null value.
+			/* type="bool" Gets/Sets whether the editor value can become null.
 				If that option is false, and editor has no value, then value is set to an empty string.
 				```
 				//Initialize
@@ -150,7 +150,7 @@
 			```
 			*/
 			inputName: null,
-			/* type="bool" Gets/Sets the readonly attribute. Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is sent into the request.
+			/* type="bool" Gets/Sets the readonly attribute for the input. If set to true the input is readonly, and all buttons and interactions are disabled. On submitting the form the editor belongs to, the value is submitted.
 			```
 				//Initialize
 				$(".selector").%%WidgetName%%({
@@ -165,7 +165,7 @@
 			```
 			*/
 			readOnly: false,
-			/* type="bool" Gets/Sets the disabled attribute.Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is not sent into the request.
+			/* type="bool" Gets/Sets the disabled attribute for the input. If set to true the input is disabled, and all buttons and interactions are disabled. On submitting the form the editor belongs to, the value is not submitted.
 			```
 				//Initialize
 				$(".selector").%%WidgetName%%({
@@ -182,7 +182,7 @@
 			disabled: false,
 			/* type="object" Gets/Sets options supported by the [igValidator](ui.igvalidator#options) widget.
 				Note: Validation rules of [igValidator](ui.igvalidator#options), such as min and max value/length are applied separately triggering errors,
-				while similar options of the editor work to prevent wrong values from being entered.
+				while the corresponding options of the editor prevent values violating the defined rules from being entered.
 				```
 					//Initialize
 					$(".selector").%%WidgetName%%({
@@ -876,7 +876,7 @@
 			}
 		},
 		field: function () {
-			/* Gets the visual editor (input) element.
+			/* Gets the input element of the editor.
 			```
 			$(".selector").%%WidgetName%%("field");
 			```
@@ -884,7 +884,7 @@
 			return this._editorInput;
 		},
 		editorContainer: function () {
-			/* Gets reference to jquery object which is used as top/outer element of th editor.
+			/* Gets a reference to the jQuery element that wraps the editor.
 			```
 			$(".selector").%%WidgetName%%("editorContainer");
 			```
@@ -892,7 +892,7 @@
 			return this._editorContainer;
 		},
 		hasFocus: function () {
-			/* Checks if editor has focus.
+			/* Gets whether the editor has focus.
 			```
 				$(".selector").%%WidgetName%%("hasFocus");
 			```
@@ -908,7 +908,7 @@
 			this._setFocusDelay(delay);
 		},
 		hide: function () {
-			/* Hides editor.
+			/* Hides the editor.
 			```
 			$(".selector").%%WidgetName%%("hide");
 			```
@@ -916,7 +916,7 @@
 			this._editorContainer.hide();
 		},
 		show: function () {
-			/* Shows editor.
+			/* Shows the editor.
 			```
 			$(".selector").%%WidgetName%%(("show");
 			```
@@ -924,7 +924,7 @@
 			this._editorContainer.show();
 		},
 		validator: function () {
-			/* Gets reference to [igValidator](ui.igvalidator) used by the editor.
+			/* Gets a reference to [igValidator](ui.igvalidator) used by the editor.
 				```
 					var validator = $(".selector").%%WidgetName%%("validator");
 				```
@@ -932,7 +932,7 @@
 			return this._validator;
 		},
 		isValid: function () {
-			/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
+			/* Checks if the value in the editor is valid. Note: This function will not trigger automatic notifications.
 				```
 				var isValid = $(".selector").%%WidgetName%%("isValid");
 				```
@@ -943,7 +943,7 @@
 			return valid;
 		},
 		validate: function () {
-			/* Triggers validation of editor and show potential warning message. If validatorOptions are set will also call validate on the [igValidator](ui.igvalidator).
+			/* Triggers validation for the editor. If validatorOptions are set will also call validate on the [igValidator](ui.igvalidator).
 				```
 				var valid = $(".selector").%%WidgetName%%("validate");
 				```
@@ -6230,7 +6230,7 @@
 			throw new Error($.ig.Editor.locale.maskEditorNoSuchMethod);
 		},
 		isValid: function () { //igMaskEditor
-			/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
+			/* Checks if the value in the editor is valid. Note: This function will not trigger automatic notifications.
 			```
 				var isValid = $(".selector").%%WidgetName%%("isValid");
 			```
@@ -9088,7 +9088,7 @@
 			return $.ui.igTextEditor.prototype.spinDownButton.call(this);
 		},
 		isValid: function () { //igDateEditor
-			/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
+			/* Checks if the value in the editor is valid. Note: This function will not trigger automatic notifications.
 			```
 				var isValid = $(".selector").%%WidgetName%%("isValid");
 			```
@@ -9670,7 +9670,7 @@
 				The following jQuery classes can be used in addition http://api.jqueryui.com/theming/icons/
 			*/
 			iconClass: "ui-icon-check",
-			/* type="number" Gets/Sets value in tabIndex for Checkbox Editor.
+			/* type="number" Gets/Sets tabIndex attribute for the editor input.
 			```
 			  //Initialize
 			  $('.selector').%%WidgetName%%({
@@ -10153,7 +10153,7 @@
 			}
 		},
 		isValid: function () { // Checkbox
-			/* Checks if value in editor is valid. Note: This function always returns true for the igCheckboxEditor.
+			/* Checks if the value in the editor is valid. Note: This function will not trigger automatic notifications.
 				```
 				var isValid = $(".selector").%%WidgetName%%("isValid");
 				```
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -38,36 +38,168 @@
 	$.widget("ui.igBaseEditor", {
 		options: {
 			/* type="string|number|null" Gets/Sets how the width of the control can be set.
+				```
+				//Initialize
+				$(".selector").%%WidgetName%%({
+					width : 200
+				});
+
+				//Get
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
 	$.widget("ui.igBaseEditor", {
 		options: {
 			/* type="string|number|null" Gets/Sets how the width of the control can be set.
```
kdinev: I don't think the option control "how" the width of the control is set. It controls the width itself.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "width", 200);
+				```
 				string The widget width can be set in pixels (px) and percentage (%).
 				number The widget width can be set as a number in pixels.
 				null type="object" will stretch to fit data, if no other widths are defined.
 			*/
 			width: null,
 			/* type="string|number|null" Gets/Sets how the height of the control can be set.
```
kdinev: I don't think the option control "how" the height of the control is set. It controls the heightitself.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Set
+				$(".selector").%%WidgetName%%("option", "height", 25);
+				```
 				string The height can be set in pixels (px) and percentage (%).
 				number The height can be set as a number in pixels.
 				null type="object" will fit the editor inside its parent container, if no other heights are defined.
 			*/
 			height: null,
-			/* type="object" Gets/Sets value in editor. The effect of setting/getting that option depends on type of editor and on dataMode options for every type of editor.*/
+			/* type="object" Gets/Sets value in editor. The effect of setting/getting that option depends on type of editor and on dataMode options for every type of editor.
```
kdinev: dataMode should be an option reference.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+			var value = $(".selector").%%WidgetName%%("option", "value");
+
+			//Set
+			$(".selector").%%WidgetName%%("option", "value", "Some text");
+
+			```
+			*/
 			value: null,
-			/* type="number" Gets/Sets value in tabIndex for editor. */
+			/* type="number" Gets/Sets value in tabIndex for editor.
```
kdinev: the value of the tabIndex attribute for the editor?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+
+			  //Get
+			  var tabIndex = $(".selector").%%WidgetName%%("option", "tabIndex");
+
+			  //Set
+			  $(".selector").%%WidgetName%%("option", "tabIndex", 1);
+			  ```
+			*/
 			tabIndex: null,
 			/* type="bool" Gets/Sets ability to prevent null value.
```
kdinev: Gets/Sets whether the editor value can become null.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Get
+				var allowNullValue = $(".selector").%%WidgetName%%("option", "allowNullValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "allowNullValue", false);
+				```
 			*/
 			allowNullValue: false,
-			/* type="string|number|null" Gets/Sets the representation of null value. In case of default the value for the input is set to null, which makes the input to hold an empty string */
+			/* type="string|number|null" Gets/Sets the representation of null value. In case of default the value for the input is set to null, which makes the input to hold an empty string
```
kdinev: What is allowNullValue is true? Description here is vague.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Get
+				var nullValue = $(".selector").%%WidgetName%%("option", "nullValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "nullValue", null);
+				```
+			*/
 			nullValue: null,
-			/* type="string" Sets the name attribute of the value input. This input is used to sent the value to the server. In case the target element is input and it has name attribute, but the developer has set the inputName option, so this option overwrites the value input and removes the attribute from the element. */
+			/* type="string" Sets the name attribute of the value input. This input is used to sent the value to the server. In case the target element is input and it has name attribute, but the developer has set the inputName option, so this option overwrites the value input and removes the attribute from the element.
```
kdinev: "This option overrides the name attribute value set to the input prior to initialization of the editor. "

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Get
+				var inputName = $(".selector").igPercentEditor("option", "inputName");
+
+			//Set
+			$(".selector").igPercentEditor("option", "inputName", "textField");
+			```
+			*/
 			inputName: null,
-			/* type="bool" Gets/Sets the readonly attribute. Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is sent into the request.*/
+			/* type="bool" Gets/Sets the readonly attribute. Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is sent into the request.
```
kdinev: "Gets/Sets the readonly attribute for the input. If set to true the input is readonly, and all buttons and interactions are disabled. On submit the value is submitted."

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Get
+				var readOnly = $(".selector").%%WidgetName%%("option", "readOnly");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "readOnly", true);
+			```
+			*/
 			readOnly: false,
-			/* type="bool" Gets/Sets the disabled attribute.Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is not sent into the request*/
+			/* type="bool" Gets/Sets the disabled attribute.Does not allow editing. Disables all the buttons and iteracitons applied. On submit the current value is not sent into the request.
```
kdinev: "Gets/Sets the disabled attribute for the input. If set to true the input is disabled, and all buttons and interactions are disabled. On submit the value is not submitted."

src/js/modules/infragistics.ui.editors.js
```diff
@@ -38,36 +38,168 @@
...
+				//Set
+				$(".selector").%%WidgetName%%("option", "disabled", true);
+			```
+			*/
 			disabled: false,
-			/* type="object" Gets/Sets options supported by the igValidator widget.
-				Note: Validation rules of igValidator, such as min and max value/length are applied separately triggering errors,
+			/* type="object" Gets/Sets options supported by the [igValidator](ui.igvalidator#options) widget.
+				Note: Validation rules of [igValidator](ui.igvalidator#options), such as min and max value/length are applied separately triggering errors,
 				while similar options of the editor work to prevent wrong values from being entered.
```
kdinev: "while the corresponding options of the editor prevent values violating the defined rules from being entered."

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+			/* Hides editor.
+			```
+			$(".selector").%%WidgetName%%("hide");
+			```
+			*/
 			this._editorContainer.hide();
 		},
 		show: function () {
-			/* Shows editor. */
+			/* Shows editor.
```
kdinev: Shows the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+			/* Set focus to the editor with delay.
+				```
+				$(".selector").%%WidgetName%%("setFocus", 200);
+				```
 				paramType="number" optional="true" The delay before focusing the editor. */
 			this._setFocusDelay(delay);
 		},
 		hide: function () {
-			/* Hides editor. */
+			/* Hides editor.
```
kdinev: Hides the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
-			/* Gets reference to jquery object which is used as top/outer element of igEditor.
+			/* Gets reference to jquery object which is used as top/outer element of th editor.
+			```
+			$(".selector").%%WidgetName%%("editorContainer");
+			```
 			returnType="$" The container editor element. */
 			return this._editorContainer;
 		},
 		hasFocus: function () {
 			/* Checks if editor has focus.
```
kdinev: Gets whether the editor has focus.

src/js/modules/infragistics.ui.editors.js
bazal4o: When the editor is cleared (by clear button on cleared and blurred) depending on this option the value of the editor is set to nullValue option, or empty string. In the derived widgets such as Numeric type there is another comment explaining the different behavior.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+			/* Gets the visual editor (input) element.
+			```
+			$(".selector").%%WidgetName%%("field");
+			```
 			returnType="$" The visual editor element. */
 			return this._editorInput;
 		},
 		editorContainer: function () {
-			/* Gets reference to jquery object which is used as top/outer element of igEditor.
+			/* Gets reference to jquery object which is used as top/outer element of th editor.
```
kdinev: Gets a reference to the jQuery element that wraps the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
 			}
 		},
 		field: function () {
-			/* Gets the visual editor element.
+			/* Gets the visual editor (input) element.
```
kdinev: Gets the input element of the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+			```
+			$(".selector").%%WidgetName%%(("show");
+			```
+			*/
 			this._editorContainer.show();
 		},
 		validator: function () {
-			/* Gets reference to igValidator used by the editor.
-				returnType="object" Returns reference to igValidator or null. */
+			/* Gets reference to [igValidator](ui.igvalidator) used by the editor.
```
kdinev: Gets a reference to...

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+				```
+					var validator = $(".selector").%%WidgetName%%("validator");
+				```
+				returnType="object" Returns reference to [igValidator](ui.igvalidator) or null. */
 			return this._validator;
 		},
 		isValid: function () {
 			/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
```
kdinev: Checks if the value in the editor is valid. 

I don't understand what the second part means, about the automatic notifications.

src/js/modules/infragistics.ui.editors.js
bazal4o: The option sets the name attribute to the hidden input which keeps the value of the editor and in case of submit its value is sent to the server. The proposed description is more general and might drive the developer to use the option in a wrong way.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -738,48 +876,77 @@
...
+				```
 				returnType="bool" Whether editor value is valid or not. */
 			this._skipMessages = true;
 			var valid = this._validateValue(this._editMode ? this.field().val() : this.value());
 			this._skipMessages = false;
 			return valid;
 		},
 		validate: function () {
-			/* Triggers validation of editor and show potential warning message. If validatorOptions are set will also call validate on the igValidator.
+			/* Triggers validation of editor and show potential warning message. If validatorOptions are set will also call validate on the [igValidator](ui.igvalidator).
```
kdinev: Triggers validation for the editor. If [validatorOptions](option_ref) are set will also call validate on the [igValidator](ui.igvalidator).

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3077,6 +3248,18 @@
 			maxValue: null,
 			/* type="bool" Gets/Sets ability to prevent null value.
```
kdinev: Same comment as the previous editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3109,7 +3292,24 @@
 				password type="string" Editor based on INPUT element with type password is created.
 				multiline type="string" multiline editor based on TEXTAREA element is created.
 			*/
-			textMode: "text"
+			textMode: "text",
+			/* type="object" Gets/Sets value in editor. The effect of setting/getting that option depends on type of editor and on dataMode options for every type of editor.
```
kdinev: the value of the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -4946,7 +5146,23 @@
 			visibleItemsCount: 5,
 			/* @Ignored@ This option is inherited from a parent widget and it's not applicable for igMaskEditor
 			*/
-			maxLength: null
+			maxLength: null,
+			/* type="object" Gets/Sets value in editor. The effect of setting/getting that option depends on type of editor and on dataMode options for every type of editor.
```
kdinev: the value of the editor.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6014,7 +6230,10 @@
 			throw new Error($.ig.Editor.locale.maskEditorNoSuchMethod);
 		},
 		isValid: function () { //igMaskEditor
-		/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
+			/* Checks if value in editor is valid. Note: This function will not trigger automatic notifications.
```
kdinev: Same as the previous editor.

src/js/modules/infragistics.ui.editors.js
bazal4o: if we want to have option reference here we need to copy paste the option and the comment in all the derived widgets, because the dataMode exists in almost every editor and we don't have a mechanism for templating the widget name within the comment, but only for the code snippet.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6040,7 +6259,20 @@
 	$.widget("ui.igDateEditor", $.ui.igMaskEditor, {
 		options: {
 			/* type="date" Gets/Sets value in editor. Date object can be set as value. String value can be passed and the editor will use the javascript Date object constructor to create date object and will use it as value. MVC date format can be used too. For example Date(/"thicks"/).
```
kdinev: the value of the editor. Date object can be set as value. String can be set and the editor will pass it to the Date object constructor and use the corresponding Date object as the value.

what is "thicks" in the last example?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6040,7 +6259,20 @@
...
+
+				//Get
+				var value = $(".selector").%%WidgetName%%("option", "value");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "value", new Date (2015, 11, 11));
+				```
+				*/
 			value: null,
 			/* type="date" Gets the minimum value which can be entered in editor by user. Date object can be set as value. String value can be passed and the editor will use the javascript Date object constructor to create date object and will use it for the comparison. MVC date format can be used too. For example Date(/"thicks"/).
```
kdinev: ...in the editor... Again use of "thicks"

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6166,6 +6398,21 @@
 			centuryThreshold: 29,
 			/* type="number" Gets/Sets difference between year in Gregorian calendar and displayed year. */
```
kdinev: This gives absolutely no description of what the option does. Is it the difference between the Gregorian calendar and the calendar in use by the application?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -9420,9 +9670,35 @@
 				The following jQuery classes can be used in addition http://api.jqueryui.com/theming/icons/
 			*/
 			iconClass: "ui-icon-check",
-			/* type="number" Gets/Sets value in tabIndex for Checkbox Editor. */
+			/* type="number" Gets/Sets value in tabIndex for Checkbox Editor.
```
kdinev: grammar.

kdinev:
Merging this one as it applies to the base editor only.

PR #309, opened 2016-09-10
Transferring code snippets for checkboxeditor

Marina-L-Stoyanova:
#295

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -9656,7 +9656,7 @@
 	});
 	$.widget("ui.igCheckboxEditor", $.ui.igBaseEditor, {
 		options: {
-			/* type="bool" Gets/Sets either the editor is checked or not.
+			/* type="bool" Gets/Sets whether the checkbox is checked.
 				```
 				//Initialize
 				$('.selector').%%WidgetName%%({
@@ -9691,7 +9691,7 @@
 				large type="string" The size of the Checkbox editor is large.
 			*/
 			size: "normal",
-			/* type="string" Applies custom class on the checkbox, so that custom image can be used.
+			/* type="string" Applies custom class on the checkbox. Custom image can be used this way.
 				The following jQuery classes can be used in addition http://api.jqueryui.com/theming/icons/
 				```
 				//Initialize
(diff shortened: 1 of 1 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -9656,10 +9656,35 @@
 	});
 	$.widget("ui.igCheckboxEditor", $.ui.igBaseEditor, {
 		options: {
-			/* type="bool" Gets/Sets either the editor is checked or not. */
+			/* type="bool" Gets/Sets either the editor is checked or not.
```
kdinev: Gets/Sets whether the checkbox is checked.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -9668,36 +9693,48 @@
 			size: "normal",
 			/* type="string" Applies custom class on the checkbox, so that custom image can be used.
```
kdinev: Applies custom class on the checkbox. Custom image can be used this way.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -9668,36 +9693,48 @@
...
 
-			  //Set
-			  $(".selector").%%WidgetName%%("option", "tabIndex", 1);
-			```
+				//Set
+				$(".selector").%%WidgetName%%("option", "tabIndex", 1);
+				```
 			*/
 			tabIndex: 0,
 			/* type="bool" Gets/Sets the readonly attribute. Does not allow editing. Disables changing the checkbox state. On submit the current value is sent into the request.
```
kdinev: "Disabled changing the checkbox state" - is this true for API calls, as well? Shouldn't I still be able to change the state programmatically?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10166,11 +10203,18 @@
 		// igCheckboxEditor public methods
 		value: function (newValue) {
 			/* Gets/Sets Current checked state/Value of the igCheckboxEditor that will be submitted by the HTML form.
-				1. If the 'value' option IS NOT defined, then 'value' method will match the checked state of the editor.
+				1. If the [value](ui.igcheckboxeditor#options:value) option IS NOT defined, then 'value' method will match the checked state of the editor.
 				This is a good option when the checkbox is intended to operate as a Boolean editor. In that case the return type is bool.
```
kdinev: This option is used when...

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10166,11 +10203,18 @@
...
-				2. If the 'value' option IS defined, then 'value' method will return the 'value' option,
+				2. If the [value](ui.igcheckboxeditor#options:value) option IS defined, then 'value' method will return the [value](ui.igcheckboxeditor#options:value) option,
```
kdinev: It won't return the value option, but it will submit the value when the editor is checked and the form is submitted.

src/js/modules/infragistics.ui.editors.js
Marina-L-Stoyanova: Yes, you can change the state programmatically. I will add this to the description.

PR #313, opened 2016-09-12
Transfer code snippets for igNumericEditor

agicheva:
#295

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -3184,52 +3184,180 @@
 
 	$.widget("ui.igNumericEditor", $.ui.igTextEditor, {
 		options: {
-			/* type="object" Gets/Sets custom regional settings for editor. If it is string, then $.ig.regional[stringValue] is assumed. */
+			/* type="object" Gets/Sets custom regional settings for editor. If it is string, then $.ig.regional[stringValue] is assumed.
+			```
+				//Initialize
+				$(".selector").%%WidgetName%%({
+					regional: "en-US"
+				});
+
+				//Get
+				var region = $(".selector").%%WidgetName%%("option", "regional");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "regional", "en-US");
+			```
+			*/
 			regional: null,
-			/* type="string" Gets the character, which is used as negative sign.
+			/* type="string" Gets/Sets the character, which is used as negative sign.
 				Note: This option has priority over possible regional settings.
-				Note: Even if the default value is null - if internationalization file is provided and it contains default values for those properties the values are imlicitly set. */
+				Note: Even if the default value is null - if internationalization file is provided and it contains default values for those properties the values are imlicitly set.
+				```
+					//Initialize
+					$(".selector").%%WidgetName%%({
+						negativeSign : "-"
+					});
+
+					//Get
+					var negativeSign = $(".selector").%%WidgetName%%("option", "negativeSign");
+
+					//Set
+					$(".selector").%%WidgetName%%("option", "negativeSign", "-");
+				```
+				*/
 			negativeSign: null,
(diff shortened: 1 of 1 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -3184,52 +3184,180 @@
...
+					//Get
+					var groupSeparator = $(".selector").%%WidgetName%%("option", "groupSeparator");
+
+					//Set
+					$(".selector").%%WidgetName%%("option", "groupSeparator", ".");
+				```
+				*/
 			groupSeparator: null,
-			/* type="array" (array of number objects) Gets the number of digits in integer part of number, which are divided into groups.
+			/* type="array" (array of number objects) Gets/Sets the number of digits in integer part of number, which are divided into groups.
```
kdinev: ... the number of digits in the integer part of a number...

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3242,9 +3370,35 @@
 				byte type="string" the Number object is used with limits of unsigned byte and if value is not set, then the null or 0 is used depending on the option 'nullable'.
 			*/
 			dataMode: "double",
-			/* type="number" Gets/Sets the minimum value which can be entered in editor by end user. */
+			/* type="number" Gets/Sets the minimum value which can be entered in editor by end user.
```
kdinev: ...in the editor by the end user.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3242,9 +3370,35 @@
...
+				//Get
+				var minValue = $(".selector").%%WidgetName%%("option", "minValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "minValue", 3);
+			```
+			*/
 			minValue: null,
-			/* type="number" Gets/Sets the maximum value which can be entered in editor by end user. */
+			/* type="number" Gets/Sets the maximum value which can be entered in editor by end user.
```
kdinev: ...in the editor by the end user.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3242,9 +3370,35 @@
...
+
+				//Get
+				var maxValue = $(".selector").%%WidgetName%%("option", "maxValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "maxValue", 100);
+			```
+			*/
 			maxValue: null,
 			/* type="bool" Gets/Sets ability to prevent null value.
```
kdinev: This comment is fixed in the base editor. Please fix it here as well.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3262,12 +3416,37 @@
 				```
 			*/
 			allowNullValue: false,
-			/* type="number" Gets/Sets delta-value which is used to increment or decrement value in editor on spin events. If value is set to negative value an exception is thrown. Non integer value is supported only for dataMode double and float.*/
+			/* type="number" Gets/Sets delta-value which is used to increment or decrement value in editor on spin events. If value is set to negative value an exception is thrown. Non integer value is supported only for dataMode double and float.
```
kdinev: what is spin events? I think the comment should explain that this happens when the spin-up and spin-down buttons are used. Also link to the option which enables the spin buttons.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -3262,12 +3416,37 @@
...
+				//Get
+				var spinDelta= $(".selector").%%WidgetName%%("option", "spinDelta");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "spinDelta", 2);
+			```
+			*/
 			spinDelta: 1,
 			/* type="null|E|e|E+|e+"
 				Gets/Sets support for E-power format in edit mode.
```
kdinev: This is called scientific format, not E-power format.

PR #327, opened 2016-09-13
Code snippets for igPercentEditor transferred.

dafo:
#295

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -5254,33 +5254,97 @@
 	});
 	$.widget("ui.igPercentEditor", $.ui.igNumericEditor, {
 		options: {
-			/* type="string" Gets the pattern for positive numeric values, which is used in display (no focus) state.
-				The "$" flag represents "numericSymbol" and the "n" flag represents the value of number.
-				Note: this option has priority over possible regional settings. */
+			/* type="string" Gets/Sets the pattern for positive numeric values, which is used in display (no focus) state.
+				If you use the "en-US" culture the default value for "positivePattern" will be "n$" where the "$" flag represents the "numericSymbol" and the "n" flag represents the value of the number.
+				Note: this option has priority over possible regional settings.
+				```
+				//Initialize
+				$(".selector").igPercentEditor({
+					positivePattern : "+ n"
+				});
+
+				//Get
+				var pattern = $(".selector").igPercentEditor("option", "positivePattern");
+
+				//Set
+				$(".selector").igPercentEditor("option", "positivePattern", "+ n");
+				```
+				*/
 			positivePattern: null,
-			/* type="string" Gets symbol, which is used in display (no focus) state.
-				Note: this option has priority over possible regional settings. */
+			/* type="string" Gets/Sets the symbol, which is used in display (no focus) state.
+				Note: this option has priority over possible regional settings.
+				```
+				//Initialize
+				$(".selector").igPercentEditor({
+					percentSymbol: "pc"
+				});
+
+				//Get
+				var percentSymbol= $(".selector").igPercentEditor("option", "percentSymbol");
+
+				//Set
+				$(".selector").igPercentEditor("option", "percentSymbol", "pc");
+				```
+				*/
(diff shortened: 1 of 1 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -5254,33 +5254,97 @@
...
+				```
+				*/
 			percentSymbol: null,
-			/* type="number" Gets/Sets the factor which used for the get and set of the "value" method.
-				On the get number (string) entered by user is divided by that factor and on the set the number (string) displayed in editor is multiplied by that factor.
-				For example, if factor is 100 and the "value" is set to 0.123, then editor will show string "12.3".
-				Possible values: 1, or 100.
-				Note: this option has priority over possible regional settings. */
+			/* type="number" Gets/Sets the factor which is used for the get and set of the "value" method.
+				On the get the number (string) entered by the user is divided by that factor and on the set the number (string) displayed in the editor is multiplied by that factor.
```
kdinev: On get... and on set

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5254,33 +5254,97 @@
...
-				uint type="string" the Number object is used with limits of unsigned int and if value is not set, then the null or 0 is used depending on the option 'nullable'.
-				short type="string" the Number object is used with limits of signed short and if value is not set, then the null or 0 is used depending on the option 'nullable'.
-				ushort type="string" the Number object is used with limits of unsigned short and if value is not set, then the null or 0 is used depending on the option 'nullable'.
-				sbyte type="string" the Number object is used with limits of signed byte and if value is not set, then the null or 0 is used depending on the option 'nullable'.
-				byte type="string" the Number object is used with limits of unsigned byte and if value is not set, then the null or 0 is used depending on the option 'nullable'.
+			/* type="double|float|long|ulong|int|uint|short|ushort|sbyte|byte" Gets the type of the value returned by the get of value() method. That also affects the functionality of the set value(val) method and the copy/paste operations of the browser.
```
kdinev: by the getter... of the setter. Add references to the value method.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5254,33 +5254,97 @@
...
+				sbyte type="string" the Number object is used with the limits of a signed byte and if the value is not set, then the null or 0 is used depending on the option [allowNullValue](ui.igpercenteditor#options:allowNullValue).
+				byte type="string" the Number object is used with the limits of an unsigned byte and if the value is not set, then the null or 0 is used depending on the option [allowNullValue](ui.igpercenteditor#options:allowNullValue).
 			*/
 			dataMode: "float", // TODO maybe it should be "double"?
-			/* type="number" Gets/Sets delta-value which is used to increment or decrement value in editor on spin events. If value is set to negative value an exception is thrown. Non integer value is supported only for dataMode double and float.*/
+			/* type="number" Gets/Sets the delta-value which is used to increment or decrement the value in the editor on spin events. If the value is set to a negative value an exception is thrown. Non integer value is supported only for dataMode double and float.
```
kdinev: Replace spin events with spinUp and spinDown. This was already done in the numeric editor, so take a look at that comment.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5505,8 +5569,8 @@
 
 		// igPercentEditor public methods
 		insert: function (string) {
-			/* Paste text at location of caret or over the current selection. Best used during editing, as the method will instead set the text as value (modified by the displayFactor) if the editor is not focused.
-				Note: method raises the "textChanged" event.
+			/* Paste text at location of the caret or over the current selection. Best used during editing, as the method will instead set the text as value (modified by the displayFactor) if the editor is not focused.
```
kdinev: `displayFactor` should be an option reference!

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5505,8 +5569,8 @@
...
+				Note: the method raises the "textChanged" event.
```
kdinev: textChanged should be an event reference!

PR #330, opened 2016-09-13
Added new message to locale files

ig-ealbert:
A new message was added in https://github.com/IgniteUI/ignite-ui/commit/cd1da6e91b835849718942439a772d9502cd6685.  I added the same message to the other locale files and added the Japanese translation we already had.

### DIFF
--- a/src/css/structure/modules/infragistics.ui.chart.css
+++ b/src/css/structure/modules/infragistics.ui.chart.css
@@ -36,7 +36,9 @@
     font-family:"Segoe UI", Arial, sans-serif;
 	position:relative;
 	background:#fff;
+	color:#666;
 }
+/*
 .ui-chart-tooltip:after {
 	top: 100%;
 	left: 50%;
@@ -50,7 +52,7 @@
 	border-top-color: #fff;
 	border-width: 6px;
 	margin-left: -6px;
-}
+}*/
 
 .ui-chart-default-tooltip-content {
     white-space: nowrap;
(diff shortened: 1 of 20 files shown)


kdinev:
@ig-ealbert I believe we have this very message translated in other localization files.

ig-ealbert:
@kdinev: Could you please specify which files have this?  I just did a search on the i18n folder for "This option can not be set runtime" and did not find any hits outside of the dialog files.  I know we had this in the docs.ja files, but the client-side APIs are not translated to any languages other than Japanese.
I found a couple of entries (dateEditorMinValue and dateEditorMaxValue) in the infragistics.ui.editors-en.js file but it doesn't completely match, so it would still need to be outsourced.

kdinev:
Here's the one in the igTree:

https://github.com/IgniteUI/ignite-ui/blob/master/src/js/modules/i18n/infragistics.ui.tree-en.js#L31

I see that there is a slight difference in the wording, so I guess it should be changed to be the same for all widgets. I'm hesitant to create a new `infragistics.ui.common-lang.js` as this will introduce one more dependency, but that's also an option in the case of common error messages for all widgets.

ig-ealbert:
@kdinev @dafo - Please advise how to proceed.  I have done the following:
1. Pulled from upstream to master.
2. Accepted @dafo's pull request on master.
3. Rebased my branch based on master.

Now I'm looking to apply the same changes to the other locale files.  I will apply the same translation as exists in the igTree locale files.  However, why was this done in the first place?  I might be going crazy, but even though my search using Notepad++ did not find any hits matching "This option can not be set runtime", I see it now in the ui.editors files.  Which means that igTree and igEditors are out of sync.  Is that ok?  Or should we undo all this and go back to the igEditors version?  And if so, should I delete my branch, close the pull request, and start over?

Also, to better understand the process in general, I'd like to know why was the pull request was made on my fork of the repo and not on the upstream master?

PR #349, opened 2016-09-16
Fix for bug 224900 - Scroll position flickers when on the bottom and scrolling down on IE.

skrustev:


skrustev:
This gets the exact width not rounded. The outerWidth() method rounds the width to integer.

kdinev:
I see. I will merge this.

Issue #370, opened 2016-09-17
Improve the API descriptions for methods


IvayloG:
These descriptions could be more meaningful, and consistent across widgets. The following is a list of the widgets to be checked for such possible improvements.

- [x]  igCheckboxEditor
- [x]  igCurrencyEditor
- [x]  igDateEditor
- [x]  igDatePicker
- [x]  igMaskEditor
- [x]  igNumericEditor
- [x]  igPercentEditor
- [x]  igTextEditor
- [x]  igBaseEditor
- [x]  igTree
- [x]  igNotifier
- [x]  igValidator
- [x]  igVideoPlayer
- [x]  igRating
- [x]  igDialog

kdinev:
@IvayloG Should #371 close this issue?

PR #371, opened 2016-09-17
Update methods' descriptions across the list of widgets.

IvayloG:
#370 

Adding updates where required.

### DIFF
--- a/src/js/modules/infragistics.ui.tree.js
+++ b/src/js/modules/infragistics.ui.tree.js
@@ -737,7 +737,7 @@
 				$(".selector").igTree({
 					dragAndDropSettings : {
 						allowDrop: true,
-						```
+						...
 					}
 				});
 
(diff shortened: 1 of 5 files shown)


src/js/modules/infragistics.ui.tree.js
```diff
@@ -747,7 +747,7 @@
 				//Set
 				settings.allowDrop = {
 					allowDrop: false,
-					```
+					...
```
kdinev: Code snippet should be copy-paste-able. Please don't leave code snippets with `...`

bazal4o: Actually only the overview code snippet is copy-paste able by our current conventions. When we leave "..." in initialize code snippet we mean there might be other options. We can replace the "..." with "//Other options go here", which will make it copy-paste able, but in that case we'll need to update most of the option's code snippets.

PR #372, opened 2016-09-17
Editors exception refactoring

agicheva:


kdinev:
@Lipata Please don't approve changes that break the build!

PR #376, opened 2016-09-19
Fix spelling errors. Add touchmove/touchend events prevention.

skrustev:


skrustev:
Currently touchstart is only getting information from the event and cancels any ongoing inertia and shouldn't cause issues but i will add it so all events of this type have the check.

PR #382, opened 2016-09-20
#380 Move method from combo.js to util.js

### LINKED ISSUE #380: Move method from combo.js to util.js
Method areArraysEqual should be added to infragistics.util.js and be referenced from there.

zdrawku:
Method areArraysEqual should be added to infragistics.util.js and be referenced from there.

Closes #380

### DIFF
--- a/src/js/modules/infragistics.util.js
+++ b/src/js/modules/infragistics.util.js
@@ -5478,20 +5478,18 @@
 	};
 
 	// Check wheather certain array of values is equal to another array
-	// This method is specifix for igCombo, and when [1, 2, 3] and [3, 2, 1]
-	// are compared, true will be returned
-	$.ig.util.areArraysEqual = function (array1, array2) {
+	$.ig.util.areSetsEqual = function (array1, array2) {
 	    var sortedArray1, sortedArray2;
 
 	    if (!array1 || !array2 || array1.length !== array2.length) { return false; }
 
 	    if (array1 === array2) { return true; }
 
-	    sortedArray1 = array1.sort();
-	    sortedArray2 = array2.sort();
+	    sortedArray1 = array1.slice().sort();
+	    sortedArray2 = array2.slice().sort();
 
 	    for (var i = 0; i < sortedArray1.length; i++) {
-	        if (sortedArray1.indexOf(sortedArray2[ i ]) !== i) { return false; }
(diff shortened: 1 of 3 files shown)


src/js/modules/infragistics.util.js
```diff
@@ -5477,6 +5477,19 @@
 		return item1 ? 1 : -1;
 	};
 
+	// Check wheather certain array of values is equal to another array
+	$.ig.util.areArraysEqual = function (array1, array2) {
+		if (!array1 || !array2) { return false; }
+
+		if (array1.length !== array2.length) { return false; }
+
```
kdinev: You can also include a check for where the two arrays are the same reference. Then there is no need to go through all the members and check them value by value.

There is also a fundamental flaw in this function, which is why unit tests need to be written. Given these two arrays:

`[1, 2, 3]` and `[1, 1, 1]` The function will return `true`

src/js/modules/infragistics.util.js
```diff
@@ -5478,16 +5478,23 @@
...
 
-		if (array1.length !== array2.length) { return false; }
+	    if (!array1 || !array2 || array1.length !== array2.length) { return false; }
 
-		for (var i = 0; i < array1.length; i++) {
-			if (array1.indexOf(array2[ i ]) === -1) { return false; }
-		}
+	    if (array1 === array2) { return true; }
+
+	    sortedArray1 = array1.sort();
```
kdinev: This will change the original arrays.

tests/unit/combo/tests.html
```diff
@@ -5505,6 +5506,77 @@ <h2 id="qunit-userAgent"></h2>
                 equal(data.length, $listItems.length, "All items are visible");
 
             });
+
+            test(testId_92, function () {
```
kdinev: Please add tests for empty arrays, and missing arguments. 

```
var x = [],  y = []; // this should yield true btw
$.ig.util.areSetsEqual();
$.ig.util.areSetsEqual(x);
$.ig.util.areSetsEqual(undefined); 
$.ig.util.areSetsEqual(undefined, y);
$.ig.util.areSetsEqual(undefined, null);
$.ig.util.areSetsEqual(null, undefined);
$.ig.util.areSetsEqual(null);      
```

zdrawku: Now everything should be okay, thanks for pointing out that more tests should be implemented.

PR #402, opened 2016-09-28
BUG-226155 // 09.28.2016 // M.Popov // [igTree] Node text gets underlined on hover

desig9stein:


src/css/structure/modules/infragistics.ui.tree.css
```diff
@@ -39,7 +39,7 @@
 	height:16px;
 }
 .ui-igtree a { 
-    text-decoration: none;
+    text-decoration: none!important;
```
kdinev: Shouldn't this be applied on `.ui-igtree a.ui-state-hover`? I believe the `!important` on the general anchor is not the best solution here.

desig9stein:
I see your point, and I believe you are right. However I saw, that
".ui-igtree a" already have "text-decoration:none" but it's overridden by a
general rule that come from "a:hover {text-decoration:none} " that's why a
give the rule important style, otherwise the "text-decoration" rule on
"ui-igtree a" is useless. I think I should remove the whole
"text-decoration" rule not just the "important" and added to ".ui-igtree
a.ui-state-hover".

kdinev:
What I meant is to remove the `text-decoration` rule from `.ui-igtree a` and to put it here:

```
.ui-igtree a.ui-state-hover, 
.ui-igtree a.ui-state-active, 
.ui-igtree a.ui-state-focus 
{
    padding:0 2px;/*compensating for the 1px borders of these other states*/
    text-decoration: none;
}
```

But that may be the same thing you're talking about :)

PR #404, opened 2016-09-28
TFS226199 adding categorychart.css for extra styling

alexkartavov:


kdinev:
@sdimchevski Can you also review the new CSS class?

sdimchevski:
Yep. Should be fine

PR #415, opened 2016-10-05
Fix for bug 115850

MayaKirova:
Adding additional logic to the $.ig.util.getRelativeOffset method so that
it can handle scenarios when the page is zoomed in.


src/js/modules/infragistics.datasource.js
```diff
@@ -8412,6 +8426,11 @@
 		},
 		updateRow: function (rowId, rowObject, autoCommit) {
 			/* updates a record in the datasource. Creates a transaction that can be committed / rolled back
+			dsMashup.updateRow(1, {
```
kdinev: shouldn't this code snippet be in a code block?

src/js/modules/infragistics.datasource.js
```diff
@@ -8359,6 +8372,7 @@
 		},
 		setCellValue: function (rowId, colId, val, autoCommit) {
 			/*  sets a cell value for the cell denoted by rowId and colId. Creates a transaction for the update operation and returns it
+			dsMashup.setCellValue(1, "Name", "CD Player", true);
```
kdinev: shouldn't this code snippet be in a code block?

kdinev:
Also please add unit tests!

kdinev:
Can you add a few more elements to the test. Currently it's just one. I just want a few more cases covered, like a table element, a list item, etc.

PR #416, opened 2016-10-05
Fix for bug 115850

MayaKirova:
Adding additional logic to the $.ig.util.getRelativeOffset method so that
it can handle scenarios when the page is zoomed in.


kdinev:
Can you add a few more elements to the test. Currently it's just one. I just want a few more cases covered, like a table element, a list item, etc. Also you can include tests with containers with different CSS positioning.

PR #426, opened 2016-10-12
Removing an old WinJS related code

kdinev:


src/js/modules/infragistics.datasource.js
```diff
@@ -25,7 +25,7 @@
  *
  */
 
-/*global define, jQuery, Class, window, ActiveXObject, DOMParser, XPathResult, XMLSerializer */
+/*global Class, ActiveXObject, DOMParser, XPathResult, XMLSerializer */
```
ChronosSF: global can be removed

PR #430, opened 2016-10-13
TFS227100 radialmenu dependency correction

alexkartavov:


src/js/modules/infragistics.util.js
```diff
@@ -5149,7 +5149,7 @@
 			obj.onTick();
 		}
 
-		if (obj.chart && obj.notify && !obj.__resizeProxy) {
+		if (obj.chart && obj.notify && obj.chart[obj.notify] && !obj.__resizeProxy) {
```
kdinev: There should be spaces in the accessor. `obj.chart[ obj.notify ]`

alexkartavov: OK, this is done, sorry about that

PR #431, opened 2016-10-13
radialmenu dependency fix

alexkartavov:


src/js/modules/infragistics.util.js
```diff
@@ -5149,7 +5149,7 @@
 			obj.onTick();
 		}
 
-		if (obj.chart && obj.notify && !obj.__resizeProxy) {
+		if (obj.chart && obj.notify && obj.chart[obj.notify] && !obj.__resizeProxy) {
```
kdinev: Same as the other PR.

PR #436, opened 2016-10-18
Filtering event returns wrong expression (master)

### LINKED ISSUE #434: [igCombo] Filtering event returns wrong expression
Filtering event in the combo does not return the current expression as `ui.expression`
object is set after the event has fired.

rkaraivanov:
closes #434

### DIFF
--- a/tests/unit/combo/tests.html
+++ b/tests/unit/combo/tests.html
@@ -328,6 +328,9 @@
     <div>
         <div id="combo-mouse-events"></div>
     </div>
+    <div>
+        <div id="filteringExpressions"></div>
+    </div>
     <script type="text/javascript">
         $(function () {
             var testId_1 = 'igCombo existence',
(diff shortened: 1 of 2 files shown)


tests/unit/combo/tests.html
```diff
@@ -5729,6 +5733,37 @@ <h2 id="qunit-userAgent"></h2>
...
+                    { "ID":3, "Name": "Bob Ferguson", "Age": 27 }
+                ];
+                $combo = $('#filteringExpressions');
+                $combo.igCombo({
+                    dataSource: data,
+                    animationShowDuration: 0,
+                    animationHideDuration: 0,
+                    valueKey: "ID",
+                    textKey: "Name",
+                    filteringType: "local",
```
kdinev: Same comment as the PR to 16.2

PR #441, opened 2016-10-19
Fix for bug 227477. Add additional tests for the igScroll.

skrustev:


src/js/modules/infragistics.ui.scroll.js
```diff
@@ -1660,49 +1673,47 @@
...
 
-			//this is to not affect the scrolling when clicking on track area of a linked scrollbarH
-			self._scrollFromSyncContentH = true;
-			this._container.scrollLeft(destX);
+			} else {
+				destX = baseElem.scrollLeft;
 
-			//}
+				//this is to not affect the scrolling when clicking on track area of a linked scrollbarH
+				self._scrollFromSyncContentH = true;
```
ChronosSF: Why mixing self and this when they are the same?

src/js/modules/infragistics.ui.scroll.js
```diff
@@ -1660,49 +1673,47 @@
...
 
-			//this is to not affect the scrolling when clicking on track area of a linked scrollbarV
-			self._scrollFromSyncContentV = true;
-			this._container.scrollTop(destY);
+			} else {
+				destY = baseElem.scrollTop;
 
-			//}
+				//this is to not affect the scrolling when clicking on track area of a linked scrollbarV
+				self._scrollFromSyncContentV = true;
```
ChronosSF: same commend regarding self and this

PR #452, opened 2016-10-25
handle exponential and circle easing functions

davidnegley:
these are already built in and just need to be handled by name


kdinev:
@davidnegley Is this applicable to 17.1 only? If not please PR to the 16.2 branch as well.

davidnegley:
17.1 only

PR #453, opened 2016-10-26
BUG-227831 // 10.26.2016 // Marin Popov //  make the icon for  "delete row"  consistent 

desig9stein:
I also fix some imperfections in the file. for example I change "0px" to "0"


kdinev:
@desig9stein Is this applicable to 16.2 as well?

PR #465, opened 2016-10-31
Fix for bug 226965: igUpload displays not a file size but a request size.

m-hristov:


m-hristov:
@kdinev, I've added test

PR #520, opened 2016-11-10
Fix for bug 228594: After updating a record in the igTreeGrid paging no longer works as expected.

m-hristov:


kdinev:
@m-hristov Pull master into your branch.

PR #532, opened 2016-11-14
Add classes for both states of the custom scrollbars so they can be customized.

skrustev:


ChronosSF:
Is this change only required for latest? If not, please make a PR to the 16.2 branch too.

Issue #460, opened 2016-10-28
Dates submitted throgh the igDateEditor/igDatePicker are always wrong


kdinev:
I have the following configuration:

```
@(Html.Infragistics()
.DatePickerFor(x => x.Session.StartTime)
.ID("StartHour")
.DateInputFormat("dd/MM/yyyy HH:mm")
.DateDisplayFormat("dd/MM/yyyy HH:mm")
.PlaceHolder("Select start hour")
.Width(280)
.Height(34)
.Render())
```

Actual behavior: When a submit a date, it gets displayed with a 2 hour offset (I'm in GMT+2). 

Expected behavior: I want to see the same time that I have submitted.

kdinev:
Apparently this works if I add ` .DataMode(DateEditorDataMode.DisplayModeText)`. If this is not the current default data mode, then it should be changed.

damyanpetev:
The DisplayModeText mode would work in this case, but display format is not necessarily required to be a valid format for the backend, so changing the default submit format (currently ISO 8601) seems a bit risky.

I see two (maybe three) issues here:
- The problem with the current toISOString, while being universally parsable, is that it's always UTC stripping the time zone info. The value is technically correct, but lacking crucial metadata and potentially also messing with usage where the timezone is not needed like this one.
- The MVC model always used the direct DateTime.toString - again no time zone to speak of, which defaults back to local time on the client. This is the actual problem causing the time shift.
- The enableUTCDates doesn't really do all ti could/should.

Ideas:
- Change the encoding of the MVC model to either similar ISO string or maybe even the MVC Json format \/Date(ticks)\/ potentially including time zone info to ensure the sameness of the value.
- However, the above might impede requirements where the client side must display the same value regardless of local zone, in which seems like a case for enableUTCDates and it could have implications on the way the date is sent from the server too. Could also dictate if the submitted value should include time zone info we can add manually (the ISO format allows for that).
- Simple text mode, kind of how it is now but with a valid universal format - think the same ISO but with DateTimeKind.Unspecified

IG-Stanimir-Dimitrov:
Fix on hold until final design (i.e. one covering new scenarios) and the corresponding API update have been agreed.

kdinev:
It also appears that with the configuration that I have, if I want to create a new date, then the seconds are actually not submitted as 00.

IG-Stanimir-Dimitrov:
igDateEditor/igDatePicker will be updated to support the following matrix:

[...]

PR #540, opened 2016-11-18
Nalipiev/date formatting

Lipata:
Please review the new descriptions of the dataMode, enableUTCDates and offset options. Review also the new messages that are thrown as exceptions. Please do not merge, cause I need to add the QUnit tests.
#460


Lipata:
Hi, @ig-ealbert. There are 3 new messages in the editors' localization files. Can you help with their translation? Thanks.

ig-ealbert:
@Lipata I will translate them for Japanese.  We outsource for the other languages once per volume release, so I will add these to the list we keep.
Do you want me to send you the translations, or submit my own pull request once this one has been merged?

Lipata:
@ig-ealbert you can send them to me. Thanks!

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7558,7 +7558,7 @@
 				$(".selector").%%WidgetName%%("option", "dataMode", "displayModeText");
 			```
 				date type="string" When this mode is set the value sent to the server on submit is serialized as UTC ISO 8061 string (e.g. "2016-11-03T14:08:08.504Z").
-					The value method returns a Date object. Both enableUTCDates and timeOffset options work only when this mode is used. That is used as default mode.
+					The value method returns a Date object. Both enableUTCDates and displayTimeOffset options work only when this mode is used. That is used as default mode.
```
kdinev: Both options should have option references!

kdinev:
@ig-ealbert Should I put a localization-pending status, or a localized status. The new string is translated to JP, but not to the other languages. I guess you do this in bulk, so the status should be `localized`, but just making sure!

Lipata:
@kdinev you are correct. I've already put the label.

ig-ealbert:
@kdinev @Lipata Yes, this label is correct because we do European translation outsourcing in bulk.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7871,11 +7894,39 @@
 		_handleSpinDownEvent: function () { // DateEditor
 			this.spinDown(1);
 		},
+		_serializeDate: function (sDate) {
+			if (this.options.dataMode === "date") {
+				if (this.options.enableUTCDates) {
```
damyanpetev: It does exactly that. The JavaScript [`Date.prototype.toISOString()`](http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.5.43) version always outputs the UTC time zone value (supposedly to simplify implementations), which is why we also use a standard polyfill (`editor._toLocalISOString()` as seen below) to produce a version with the local zone.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10396,15 +10411,17 @@
...
 		},
 		getSelectedDate: function() {
-			/* Gets selected date.
+			/* Gets selected date as a date object. This method can be used when dataMode is set as either displayModeText or editModeText.
+			In such cases the value() method will not return date object and getSelectedDate() can be used to replace that functionality.
 			```
 			$(".selector").%%WidgetName%%("getSelectedDate");
 			```
 				returnType="date" */
 			return this._dateObjectValue;
```
damyanpetev: That'd be the internal date object the editor is based around, if it's wrong at any given point something must've gone terribly wrong elsewhere. Could re-create the object (from the internal one) purely to avoid unintentional changes to the return value from applying to the editor. There's `selectDate` for that after all :)

kdinev:
Can you add this to the test cases?

http://ideas.infragistics.com/forums/211535-ignite-ui/suggestions/13428417-ig-gridexcelexporter-export-utc-dates-with-the-di

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7736,6 +7747,9 @@
 
 			// RegEx for /Date(milisecond)/
 			this._mvcDateRegex = /^\/Date\((.*?)\)\/$/i;
+			if (offset !== null && (offset > 840 || offset < -720)) {
+				console.log($.ig.Editor.locale.dateEditorOffsetRange);
```
IG-Stanimir-Dimitrov: @Lipata , I discussed this with @kdinev. 
Can we throw a configuration error if displayTimeOffset is outside the valid boundaries?

damyanpetev: Fair enough, I'll change this case and I've added #681 for other instances to be changed to follow what the majority of controls do with incorrect options.

damyanpetev:
@kdinev I've made changes according to the comments above, except the case of setting `displayTimeOffset` while in edit mode. Was for it initially, but exiting edit mode actually requires also blurring the editor which seems way too disruptive.

As for the exporter idea - that's purely based on the exporter and data source as far as I can tell, so I can't really cover it in this PR/issue.

kdinev:
I don't think the coverage should go down with this change.

damyanpetev:
@kdinev I've deleted mostly lines that were covered (replacing them with just one) so the covered count vs the total relevant ones should be relatively lower, no?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10750,12 +10720,21 @@
 					} else {
 						date = self._setNewDateMidnight();
 					}
-					date = self._setDateField("year", date, dateFromPicker.getFullYear());
 
-					//Temporary change the date to be in the middle of the month 15th, because when using JavaScript Date object to set month when date is 31, the date object is moved with one day.
-					date = self._setDateField("date", date, 15);
-					date = self._setDateField("month", date, dateFromPicker.getMonth());
-					date = self._setDateField("date", date, dateFromPicker.getDate());
+					if (self.options.displayTimeOffset !== null) {
```
kdinev: I checked the coverage and the date picker is tested only when `displayTimeOffset` is default (`null`). Please add date picker tests with different values for the `displayTimeOffset`

PR #542, opened 2016-11-21
Fix for bug 228717. Scrollbars not updated when changing scrollWidth/Height. Add tests.

skrustev:


tests/unit/scroll/tests.html
```diff
@@ -3475,6 +3478,66 @@
 				start();
 			}, 2500);
 	    });
+		
+		test(testId_126, function () {
+	        $("#testBed").append($('<div id="elemVH" style="height:400px; width: 600px; overflow: hidden;" tabindex="1"></div>').append(contentScrollVH));
```
kdinev: You can cache the newly created element here and reuse it as a cached reference down the test, instead of selecting it by ID. There will also no longer be need to put ID on it.

tests/unit/scroll/tests.html
```diff
@@ -3475,6 +3478,66 @@
...
+			$('#elemVH').igScroll("option", "scrollWidth", 500);
+			
+			equal($("#elemVH_hBar").length, 0, "Horizontal thumb drag is removed");
+			equal($("#elemVH_vBar").length, 1, "Vertical thumb drag is not removed");
+
+	        $("#elemVH").remove();
+	    });
+		
+		test(testId_127, function () {
+	        $("#testBed").append($('<div id="elemVH" style="height:400px; width: 600px; overflow: hidden;" tabindex="1"></div>').append(contentScrollVH));
```
kdinev: Same comment as above.

PR #544, opened 2016-11-24
igCombo unable to select html entity in datasource

### LINKED ISSUE #543: [igCombo] Cannot select an item from drop down list when you set HTML character entity reference to a datasource item.
When HTML character entity reference is set to datasource, it is escaped regarding drop down item text.
But it isn't escaped regarding data-value attribute.

For example
`<li data-value="¥">&amp;yen;</li>` should be `<li data-value="&amp;yen;">&amp;yen;</li>`

rkaraivanov:
closes #543


kdinev:
@rkaraivanov Can you also include a test in the templating that verifies that ampersands are encoded? There is a test for encoding already, just include the ampersand in it.

PR #547, opened 2016-11-24
Deleting the code mentioned as unused.

### LINKED ISSUE #503: Unreachable code in the public value method for igNumericEditor
The following is the relevant code snippet:

      if (newValue < this.options.minValue) {
       newValue = this.options.minValue;
      } else if (newValue > this.options.maxValue) {
       newValue = this.options.maxValue;
      }
      this._updateValue(newValue);
      if (!this._focused) {
      } else {
       this._enterEditMode();
      }

IvayloG:
minValue and maxValue are correctly handled for the valid numbers input, and
NaN scenarios are handled to either clear the invalid NaN or revert to last valid one base on revertIfNotValid option.

closes #503

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -5557,18 +5557,6 @@
 						} else {
 							this._enterEditMode();
 						}
-					} else {
-						if (newValue < this.options.minValue) {
-							newValue = this.options.minValue;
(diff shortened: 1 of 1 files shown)


IvayloG:
Actually it seems related to a case when a list with values is used. I will further look into this matter.

IvayloG:
After further looking into the matter it appears the code does not need to be refactored as it covers real use case scenarios.

Issue #558, opened 2016-11-28
Unclear and/or wrong descriptions in igToolbarButton API


hanastasov:
1) the description of isSelected option is wrong
2) activated/activating/deactivate/deactivating description says:
Use ui.owner to get reference to the toolbar button performing the deactivation.
It sounds better just:
Use ui.owner to get reference to this igToolbarButton
3) No description is available for http://www.igniteui.com/help/api/2016.2/ui.igtoolbarbutton#theming:ui-igtoolbarbutton

Issue #559, opened 2016-11-28
Unclear and/or wrong descriptions in igToolbar API


hanastasov:
1) http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#options:allowCollapsing
2) It is not clear which is the expandButtonIcon http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#options:expandButtonIcon
- better be smth like "the expand/collapse button icon"
3) http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#options:isExpanded
- needs better description
4) http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#events:collapsing
- desc is wrong !
5) http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#events:toolbarComboOpening
- needs better desc
6) http://www.igniteui.com/help/api/2016.2/ui.igtoolbar#methods:getItem make the descriptiong consistent with the removeItem method

kdinev:
Should I close this issue?

IG-Stanimir-Dimitrov:
Approved by PO.
Changes were merged with pull request - fix description in igToolbarButton, igToolbar and igTileManager APIs #608

PR #560, opened 2016-11-28
fix description to igToolbarButton API

Aleksandyr:
#558


Aleksandyr:
This commit automatically was associated with this pull request. It had to be another pull request that is associated to task #559

Issue #539, opened 2016-11-18
[igNumericEditor] If min/max value is set to 0 and the entered value is invalid, the editor's value is not reverted


agicheva:
**Steps to reproduce**
Create a sample with the following code:
```
$("#editor").igNumericEditor({
    maxValue: 10,
    minValue: 0
});
```
Enter a negative value and focus out the editor

**Actual result**
The negative value is set to the editor

**Expected result**
The value of the editor is supposed to be reverted to the min value

**Comment**
It is reproducible if the maxValue is set to zero and the entered value is positive.

Issue #487, opened 2016-11-04
[igBaseEditor] input value property is not restored when destroy method is called 


IvayloG:
If there is a text in the input prior to initializing the widget it is not restored when calling widget's destroy method.
This happens because input's value attribute and value property are decoupled after the initial input rendering.

Issue #446, opened 2016-10-21
TextEditor doesn't trigger valueChange after clearing


agicheva:
**Steps to reproduce:**
Create a sample with the following code: 

```
$("#firstName").igTextEditor({
                buttonType : "clear",
                valueChanged: function(evt, ui){
                    console.log("Trigger valueChanged");
                }
            });
```

Enter text in the editor and focus it;
Click on the "clear" button;
Focus out the editor;

**Actual result:**
The valueChanged event is not triggered and the "Trigger valueChanged" is not logged in the console.

**Expected Result:**
The valueChanged event is supposed to be triggered and in the console is supposed to be logged "Trigger valueChanged".

**Comment:**
Test it with KnockoutJS.
http://www.igniteui.com/editors/bind-editors-with-ko

PR #567, opened 2016-11-29
IGanchev/igNumericEditor min/maxValue fix16.2

IvayloG:


kdinev:
Are #446 and #487 only applicable to 17.1+?

IvayloG:
Yes they are. Only #539 test should make it to 16.2.

Issue #448, opened 2016-10-24
Suppres keyboard showing when multipleSelection is used


IG-Stanimir-Dimitrov:
**Client requirement**
Not focus the combo and have the keyboard pop up on touch devices when the SuppressKeyboard flag is on to allow the user to keep tapping all the options

kdinev:
Duplicate of #284 
Already done.

zdrawku:
When suppressKeyboard flag is on along with multipleSelection, the focus should not move to the combo input in order keyboard to remain hidden while choosing different items from the dropdown list.

PR #580, opened 2016-12-01
[igCombo] Merge suppressKeyboard for multiple selection in 16.2

rkaraivanov:
Push the changes for #448 into 16.2 as well.
Also pushed the commit from @MayaKirova for the date specific test as it is not commited to 16.2


TammyD79:
Is there an ETA on this? I thought it would be in the SR this week but it's not. I didn't realize my github entry was merged into this one...

PR #592, opened 2016-12-02
Format not fully formatted date, when pasted in the editor.

### LINKED ISSUE #504: [igDateEditor] Copy date as plain text and paste it in the editor is not setting value
**Steps to reproduce:**
Create a code sample with the following code: 
```
$("#dateEditor").igDateEditor({
                      width: 280,
                      dateDisplayFormat: "dateTime",
                      dateInputFormat: "dateTime"
                  });
```
Copy this date: 11/3/2015 3:24 PM
Paste it in the editor

**Actual result:**
The value is reverted

**Expected result:**
The value of the editor to be 11/3/2015 3:24 PM

Lipata:
Fixes #504

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -7983,6 +7983,67 @@
 		_getInternalMaskedValue: function (newDate) {
 			return this._updateMaskedValue(newDate, true);
 		},
+		_replaceDisplayValue: function(selection, previousValue, newValue) {
+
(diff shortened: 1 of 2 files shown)


IvayloG:
A couple of the code comments could be optimized:
 // This method is used only for date editor/picker to transform not fully **fully** formatted dates, like 1/3/2015 3:24 PM, to 11/_3/2015 _3:24 PM.
// In addition, in the date editor, we need **to** format this to be correct date.

PR #594, opened 2016-12-02
Fix for issue with adding new records with paging and grouping in angular scenario

MayaKirova:
This is related to bug 228845.


kdinev:
@MayaKirova please include the test in this PR as well.

PR #602, opened 2016-12-05
Force the position of the igPopover when a direction is specified

dkamburov:


kdinev:
@dkamburov Is this tested with the rest of the directions? I see that only direction right is added in the tests. Does this mean direction left, and bottom are covered by already existing tests?

dkamburov:
I'll add more test to cover all the directions, but first let's clear out the approach for this issue to be fixed, it may require another approach

PR #603, opened 2016-12-05
Force the position of the igPopover when a direction is specified

dkamburov:


dkamburov:
Well, I agree about revising _forced property and also revising positioning logic of the popover, but at this point _forced can be set [based on the result of the method _findProperPosition](https://github.com/dkamburov/ignite-ui/blob/22cd705706440a88ca2e314fd7256b044036117c/src/js/modules/infragistics.ui.popover.js#L624)

ChronosSF:
Sure. It may be worth annotating where this comes from though, since it looks like a hack to get the next _checkCollision to not return false (basically making the priority cycle pointless, since it should always succeed on the first try, at least when it comes to _checkCollision).

kdinev:
@dkamburov Is this tested with the rest of the directions? I see that only direction right is added in the tests. Does this mean direction left, and bottom are covered by already existing tests?

dkamburov:
I'll add more test to cover all the directions, but first let's clear out the approach for this issue to be fixed, it may require another approach

PR #631, opened 2016-12-08
Amd modules

alexkartavov:


kdinev:
Nooooo :) I will close this one because it merges the entire master into 16.2. Could you cherrypick just your change?

PR #634, opened 2016-12-09
When immediate mode is enabled, then fire valueChanged event properly.

Lipata:


Marina-L-Stoyanova:
There is one case in which I'm able to reproduce the issue: Focus the igTextEditor and click on the clear button. The valueChanged event is not triggered.

Marina-L-Stoyanova:
The issue turned out to be reproducible only in 16.2.

Issue #577, opened 2016-11-30
igTextEditor valueChanged event is not firing when the value in the editor is changed


igMPeterson:
igTextEditor valueChanged event is not firing when the value in the editor is changed. Run the attached sample to see the behavior. 
[igTextEditorKnockout.zip](https://github.com/IgniteUI/ignite-ui/files/622548/igTextEditorKnockout.zip)

TammyD79:
Is there an ETA on this?

kdinev:
@TammyD79 We're starting a new sprint on Monday and we will put it in that sprint. ETA would be December 19th the latest.

PR #635, opened 2016-12-09
When immediate mode is enabled, then fire valueChanged event properly.

Lipata:
#577


Marina-L-Stoyanova:
There is one case in which I'm able to reproduce the issue: Focus the igTextEditor and click on the clear button. The valueChanged event is not triggered.

PR #679, opened 2016-12-19
Fix for bug #230194

tbelcheva:
Missing return type of igDataSource pageSize and pagingSettings API
methods


src/js/modules/infragistics.datasource.js
```diff
@@ -2243,6 +2243,7 @@
 				var pagingSettings = ds.pagingSettings();
 			```
 			paramType="object" optional="true" object holding all paging settings. See settings.paging
+			returnType="object"
```
kdinev: Add a comment that the instance of the data source is returned.

Issue #646, opened 2016-12-12
In igNumericEditor you can set the value to null on run time when allowNullValue=false


Marina-L-Stoyanova:
Create a sample that has set allowNullValue to null. Try to set the value to null run time through the option and then get the value again through the option. The returned value is null. It shouldn't be.

kdinev:
@Marina-L-Stoyanova You can actually attach a sample to the issue :)

Marina-L-Stoyanova:
Sure here is the sample. 
[igNumericEditor-sample.zip](https://github.com/IgniteUI/ignite-ui/files/648991/igNumericEditor-sample.zip)
**Steps to reproduce:**
1. Open the attached sample.
2. enter a value for example 10.
3. Click the set null button.
4. Open the console and write $('#text').igNumericEditor("option","value")

**Actual result:**
The returned value is null

**Expected result:**
The returned value shouldn't be null as the allowNullValue option is set to false.

PR #685, opened 2016-12-20
Fix igNumericEditor value setting using value option.

bazal4o:
#646


kdinev:
@MayaKirova There seems to be an issue with one of the Date related tests of the date source. Could you please take a look?

PR #686, opened 2016-12-20
Tpaskalev/issue 646

bazal4o:


tests/unit/editors/Bugs/tests.html
```diff
@@ -1235,6 +1243,20 @@
 
 				$("#539Editor").remove();
 			});	
+			testId = 'Bug 646 - Set null value using setOption';
+			test(testId, 2, function () {	
+				var editor =  $("#646Editor"),
+				field = editor.igNumericEditor("field");
+				equal(editor.igNumericEditor("value"), 20, "value should be 10");
```
kdinev: I think the error message should say the value should be 20.

PR #687, opened 2016-12-21
Fix for bug 230198

kacheshmarova:
Fix for bug 230198 version 17.1


kdinev:
Just fix the code style error.

PR #688, opened 2016-12-21
Fix for bug 230198

kacheshmarova:
Fix for bug 230198 Version 16.2


kdinev:
Just fix the code style error.

PR #691, opened 2016-12-21
Fixing issues in API for 17.1

tbelcheva:
#230188 - Not complete code snippet
#230196 - Removing internal parameters


kdinev:
@tbelcheva this should be pulled into 16.2 as well.

Issue #696, opened 2017-01-03
[igCombo] Combo doesn't accept first character of input after clearing an IME composition in Chrome


rkaraivanov:
After clearing the combo input and turning on direct input (A), the first character entered is composited again.

rkaraivanov:
closed via https://github.com/IgniteUI/ignite-ui/pull/702

PR #702, opened 2017-01-04
#696 [igCombo] Combo doesn't accept first character of input after clearing an IME composition in Chrome 

rkaraivanov:
Left only special composing for Safari. Should investigate if Safari on macOS is according to standard so that we can remove the whole chunk of special handling code.


rkaraivanov:
The fix for #709 ended up in this PR as well. 
@hanastasov can you test this too

Issue #693, opened 2016-12-29
[htmlEditor] - Error occurs in IE11 when inserting image using insert object toolbar.


petko5kov:
"SCRIPT5007: Unable to get value of the property 'nodeType': object is null or undefined" occurs in IE11 when inserting image using insert object toolbar.

PR #703, opened 2017-01-04
[16.2] Fix to #693; Test is created

petko5kov:


kdinev:
@petko5kov Could you also PR to master?

PR #708, opened 2017-01-05
Added message for commercial support

ig-ealbert:
This message was only added in English previously.  I added it in Japanese (translated) and to the other locale files in English.


kdinev:
@ig-ealbert is this also applicable to 16.2?

ig-ealbert:
@kdinev I'm not sure but I think so?  I'm just following the changes made in the i18n folder.  Should I be checking "master" or "16.2" for updates?

ig-ealbert:
@kdinev It doesn't look like this was applied in 16.2 in English.  If that was intentional, then it only applies to 17.1 and we're done.

PR #721, opened 2017-01-09
#720 Fixed hiding header and footer in igLayoutManager

### LINKED ISSUE #720: [IgLayoutManager] Hiding header and footer does not work
Open http://jsfiddle.net/7g82ged8/1/
The header, footer and right sidebar are shown.
The header, footer and right sidebar should be hidden.

tiliev:
Closes #720


kdinev:
@hanastasov Should this be returned to `status: verified`?

PR #723, opened 2017-01-10
Fix igScroll not being updated when width/height are set in % and the window is resized.

### LINKED ISSUE #229: [igScroll] does not update if the container has width in % and the window is resized
If the horizontal scrollbar is visible, the container with is set in % and the width changes when the window is resized, then the horizontal scrollbar width should also change with it.

skrustev:
Resolves #229 
[Sample](https://github.com/IgniteUI/ignite-ui/files/695672/igScroll-percentage.gz)

Steps to reproduce:
1. Run the attached sample
2. Resize the window
3. Resize the "#table-div" element by changing it's width or height

Actual result:
The igScroll did not update it's containers and the scrollbars didn't resize as well.

Expected result:
The igScroll containers and scrollbars should resize accordingly.


mpavlinov:
Test suggestion: Put the scroll in an igSplitter container and resize the panels to see if it works.

mpavlinov:
@kdinev: I re-requested your review, because of an additional commit after your first review. In the new commit 2 new events are introduced: "resizing" and "resized" and [MutationObserver](https://developer.mozilla.org/en/docs/Web/API/MutationObserver) is used to detect the container dimension changes.l

kdinev:
@mpavlinov Yeah, I saw that. I dismissed my previous review and added a new one. I'm fine with the events and with the usage of `MutationObserver`.

PR #724, opened 2017-01-10
#701 Spin properly date editor value

### LINKED ISSUE #701: [igDateEditor] Spin delta not always respected - buttons + API
1. Open sample from [igDateEditor-spin.zip](https://github.com/IgniteUI/ignite-ui/files/682539/igDateEditor-spin.zip)
2. Press an editor spin button
3. Press the external spin button

**Result:** 
The first time the value is modified by 1 instead of the defined `spinDelta` of 2. Same happens from the API call (2nd button).

**Expected:**
Both built-in buttons and API methods should default to the delta option and the API method should not be triggering an event.

**Note:** This is a result of the button handles like `_handleSpinUpEvent` calling directly the public method with set "1" as parameter. A side effect of the logic going through one method is that everything triggers and event as well. Will need some refactoring and while at it - the description on `spinDelta` for the date editors needs a review.

Lipata:
Spin properly date editor value, when spinDelta defined. Fire propery textChanged and valueChanged events.

closes #701


damyanpetev:
@Lipata Can you also add the same exception when the `spinDelta` is negative on initialization like the numeric editors.

Lipata:
@damyanpetev In addition I have added check when spinDelta is not a number and parse to int double and float values.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10289,13 +10308,14 @@
 			if (!date) {
 				date = this._setNewDateMidnight();
 			}
-			if (newPeriod !== period) {
-				this._setDateField(periodName, date, newPeriod);
+			this._setDateField(periodName, date, newPeriod);
+			this._editorInput.val(this._getDisplayValue());
+			if (userInteraction) {
```
damyanpetev: Actually, moving value processing (`_triggerInternalValueChange`) to optional step later can leave `_getDisplayValue()` applying the current value before the spin (e.g. empty initial value will get updated to today, but the display will be left blank).

PR #729, opened 2017-01-11
#695 Changing the value parsing method  

### LINKED ISSUE #695: [igPercentEditor] Focusing the widget causes it's value to be multiplied by 10000 when using regional "de-DE"
**Steps to reproduce**
Open the following sample:
http://jsfiddle.net/j940dg6s/
Focus/Click the igPercentEditor on the bottom **D**

**Actual result**
In editing mode it displays 123400
After blur it displays 123.400,00%

**Expected result**
In editing mode it should display 12,34
After blur it should display 12,34%

IvayloG:
Using parseFloat instead of the internal _parseNumericValueByMode

closes #695


kdinev:
I believe that the change should be in our parser, instead of switching to `parseFloat`. The parser logic should be updated to take into account the specific separators for the culture currently being used.

damyanpetev:
@kdinev The change affects parsing of value coming from the internal (hidden) input that only gets assigned an actual parsed Number so the native parse should be able to handle this (without needing to worry of symbols and separators or rounding due to maxDecimals). Frankly, the value in options could replace this as well (with a null/NaN).

dafo:
The described scenario is fixed but the same issue is causing another misbehavior.
Steps to reproduce:
1. Navigate to http://jsfiddle.net/j940dg6s/
2. Click inside the last plain input ("D"). 
3. Change the value from 0.1234567**89** to 0.1234567**90**
4. Blur the editor
5. Take a look at the igPercentEditor ("D")

Actual Result: 
The value is changed to 12.345.679.000,00%

Expected result:
The value should be correct

Note: The same can be reproduced in the igNumericEditor ("A" and "B")

damyanpetev:
@dafo Right that'd be the parsing the the value method so applicable to all numeric editors. Same unexpected result due to the value coming from the model update being in the JS number format and the '.' ignored as group separator.
There's already a NaN check in the extensions, so maybe we can also add a parse there?
Maybe also document the string parameter side of the value method?

dafo:
The original scenario and the scenario that I`ve described earlier are working as expected now. But the fix is causing another issue:

**Steps to reproduce:**
    1. Navigate to http://jsfiddle.net/j940dg6s/
    2. Click inside the last plain input ("D").
    3. Delete the value
    4. Blur the editor
    5. Take a look at the igPercentEditor ("D")

**Actual Result:**
The value is reverted to 12,34%

**Expected result:**
The values should be the same (both editor shopuld be empty)

**Note**: The same can be reproduced in the igNumericEditor ("A" and "B")

PR #742, opened 2017-01-17
Fix for bug #229997

tbelcheva:
When paging is enabled after editing a value in TreeGrid the tooltip
shows incorrect text


kdinev:
I remember this piece of code was added as a bug fix for another bug. Has that bug been re-verified?

MayaKirova:
this.settings.type is not always populated so this would cause the grid to rebind in scenarios when it shouldn't. You can use this._runtimeType, which gets populated on dataBind.
I've attached a sample to the related bug.

PR #750, opened 2017-01-18
Combo allowCustomValue for KO bindings

### LINKED ISSUE #746: [igCombo] Custom values are not persisted in the combo input
### Description  
When using the Knockout bindings, having `allowCustomValue` set to `true` do not persist the custom value in the combo input.

### Steps to reproduce  

1. Run the attached sample.
2. In the Knockout Combo, delete all the text and type "Bl" and then "o".

### Result  
When typing an unmatched item, the text is cleared from the combo's input.

### Expected result  
The text should remain in the input when `allowCustomValue` is set.

### Attachments  

[igComboKO_CustomValues.zip](https://github.com/IgniteUI/ignite-ui/files/711555/igComboKO_CustomValues.zip)

rkaraivanov:
Closes #746


tests/unit/knockout/combo/tests.html
```diff
@@ -332,13 +348,24 @@
 			$element.trigger(mouseUp);
 		}
 
+		function typeInInput(key, element) {
```
kdinev: These methods should be in the `TestUtil`. I believe there is already a method there defined for the same thing.

Issue #656, opened 2016-12-13
igNotifier is not properly positioned when there is not enough space


Marina-L-Stoyanova:
**Steps to reproduce:**
Open the attached sample.
Focus the editor. 
Look at the notifier.

**Actual result:**
Half of the message is hidden, because its position is balanced to the editor.

**Expected result:**
The message should be properly aligned to the editor if there is no enough space.

[igNotifier.zip](https://github.com/IgniteUI/ignite-ui/files/649169/igNotifier.zip)

PR #754, opened 2017-01-19
Popover improvements #637, #656

### LINKED ISSUE #637: When the direction of the popover is left, it should show up from that direction, even though there is not enough space
[popover.zip](https://github.com/IgniteUI/ignite-ui/files/642610/popover.zip)

ChronosSF:
Closes #637 , #656  .  

Exposing direction priority as an option. Enforcing direction when user-specified.


src/js/modules/infragistics.ui.popover.js
```diff
@@ -162,38 +166,38 @@
...
+				"left":"right"
+			};
 			this._positions = [ "balanced", "start", "end" ];
-			this._directionIndex = -1;
-			this._positionIndex = -1;
 			this._visible = false;
 			$( window ).on( "resize.popover", $.proxy( this._resizeHandler, this ) );
 		},
 		_createWidget: function (options, element) {
+			// initialization performance will be better if we don't try to normalize the defaults
```
kdinev: Options should not be cached in an internal property. If you're changing the value of the option, then assign it back to the option. This causes a lot of issues long term.

tests/unit/popover/common/tests.html
```diff
@@ -473,7 +473,8 @@
                 ok( $( '#notifier_popover_contentFrame' ).css( "display" ) !== "none", 'Popover content not shown!' );
                 ok( $( '#notifier_popover_arrow' ).css( "display" ) !== "none", 'Popover arrow not shown!' );
                 ok( $( '#notifier_popover_arrow' ).hasClass( "ui-igpopover-arrow-bottom" ), 'Popover arrow is not pointing to bottom!' );
-                ok( $( '#notifier_popover' ).offset().top + $( '#notifier_popover' ).height() <= $( '#notifier' ).offset().top,
+                ok( $( '#notifier_popover' ).offset().top + $( '#notifier_popover' ).height() <= $( '#notifier' ).offset().top ||
```
kdinev: There is quite a few changes in the logic, and only one modification in the unit tests. Please write exhaustive unit tests for the new changes.

ChronosSF:
@kdinev , I removed the priority cache and instead I am assigning back the "fixed" prio array to the option. Also added tests unique to the new option. The other logic changes are pretty well covered by the remaining tests (had to change a few due to the new overlap enforcement in some cases).

PR #755, opened 2017-01-19
Fix for bug #229440

tbelcheva:
When new rows are added to the grid while there are grouped columns and
paging is enabled the rendered rows count does not match the total count
displayed in the pager label


kdinev:
@MayaKirova @tbelcheva I remember the line that is removed with this PR was added by Miro for a bug fix where duplicate rows were added in the TreeGrid. Could you please verify if that bug is reproducible again?

PR #762, opened 2017-01-24
Initial date editor min/max setting

### LINKED ISSUE #731: [igDateEditor] Setting value larger than the maximum date doesn't set the latter as editor value
The following configuration will not set editor's value to the maximum date - 5 December 2016:

```
$('#edtr').igDateEditor({
	value: new Date(2016, 11, 10),
	maxValue: new Date(2016, 11, 5),
	minValue: new Date(2016, 11, 1),
	buttonType: "spin"
});
```

Lipata:
Closes #731 

If initial editor value exceeds the min/max value, then set it to min/max and show notification. Correct the minimum value warning message.


Lipata:
@ig-ealbert can you check and confirm that the Japanese message minValExceedSetErrMsg was correct, while in all the other languages it was wrong.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8974,6 +8969,23 @@
 			}
 			return result;
 		},
+		_validateMinMax: function(date) {
```
bazal4o: rename the method to _getValueBetweenMinMax, or similar, because it's not returning Boolean value -what is expected from the name.

Lipata: It's renamed. Thank you!

ig-ealbert:
@Lipata it is correct in Japanese.
入力値が最小値である {0} より小さいため、最小値に設定されました。
より小さい means "less than".

PR #763, opened 2017-01-24
Numeric/Text list spin

### LINKED ISSUE #657: [igPercentEditor] The value is incremented by 1 when the isLimitedToListValues is set to true
**Steps to reproduce**
Open the following sample:
http://jsfiddle.net/8j5wxywu/
Select value from the dropdown and spin up/down 

**Actual result**
The value is incremented/decremented by 1.

**Expected result**
The value is supposed to be the previous/next element from the listItems. 

**Comment**
It is reproducible for igNumericEditor and igPercentEditor.

damyanpetev:
Closes #657 

Starting this PR for early reviews and discussion.
After reviewing the current list support with @IG-Stanimir-Dimitrov this became more of new functionality than a bug, since we wanted text and numeric to be fairly consistent.

So far (outside of bunch of refactoring):
- Added logic to keep list selected item in sync with values and edits.
- Text editor spin buttons now actually select and work regardless of dropdown state. (the old hover was kind of pointless if you have to go click the item again)
- Updated spin button state for text editor to actually work as well
- Multiple public methods updated (replaced hover for select), `selectedListIndex` does what you'd expect now and ` selectListIndexUp/Down` for numeric editors are deprecated as they performed spin anyway.
- Numeric editor with `isLimitedToListValues` set now transfers most of it's spin functionality to the text editor.
- Since text editor already allowed keyboard navigation with enabled read-only dropdown, the same is now true for numeric. I'm hoping this covers the usage in the grid page size selector (without auto show list ofc.) as it wasn't so far.

I still want to update the numeric/percent spin a bit more and will be adding more tests for sure. I've kept keyboard navigation behavior intact even though it's a bit different for [numeric](http://www.igniteui.com/help/ignumericeditor-keyboard-navigation#ddInteraction) and [text](http://www.igniteui.com/help/igtexteditor-keyboard-navigation#ddInteraction), but I think we don't want to be changing that one.

Questions: 
- I see the original `isLimitedToListValues` validation was performed case-insensitive. Should that be the case for list matching too?
- As this is supposed to be a very basic dropdown list so we let the default editor behavior handle the value, so there are chances the selected/spin item will be rejected or reverted (`revertIfNotValid`). Don't really want to be doing more than that for conflicting options, right?


damyanpetev:
Alright, after the last changes and refactoring of numeric spin methods (bai bai coverage..) this should be ready for testing.

kdinev:
@damyanpetev @bazal4o There is a pending review on Todor. I guess I should wait for the review before merging?

damyanpetev:
Discussed with @IG-Stanimir-Dimitrov and we are keeping the case-insensitive match for value and that also applies to selection. Also stripping any values the list can't display from the array.
Per @ig-ealbert's comment on #770 she'll update the API changes in bulk, so there should be nothing left outstanding on this PR.

ig-ealbert:
Yes, we just got the 17.1 API build set up, so I can start working on that this week.

PR #770, opened 2017-01-26
#626 Round decimal places by default

### LINKED ISSUE #626: [igNumericEditor] When the value exceeds maxDecimals length, the editor directly truncates it
The igNumericEditor currently always truncates the decimal positions overflowing the maxDecimals length. The expected behavior for the editor is to round up to the nearest number in such cases.

Changes:

- The editor must round the value up to the max decimal precision (default behavior)

- Implement an option that sets whether values, which precision exceeds the maxDecimals length, are truncated or rounded (enhancement)

Lipata:
Closes #626  

Round values, when decimal places are more than the allowed, set at the maxDecimals option.


src/js/modules/infragistics.ui.editors.js
```diff
@@ -4808,6 +4840,9 @@
 			}
 			return x;
 		},
+		_roundExponential: function(value, decimals) {
```
damyanpetev: We discussed this is done due to `toFixed()` rounding, however I found the decimal rounds like [`Math.round10`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round#Decimal_rounding) in our util file - seems to do exactly what you need, so maybe use that instead? :)
P.S. Can you also copy the description from MDN and add a link where it came from in util?

Lipata: Done. Thanks!

damyanpetev:
Rounding now behaves correctly and as far as I can tell in line with the rounding of the old editors. Speaking of:
- @Lipata Can we update the `maxDecimals` description to ".. decimal places ~~which are used in display mode(no focus)~~ supported by the editor". That seems to be a leftover from the old editor (where it was wrong as well mind you).
- @IG-Stanimir-Dimitrov Should this be documented as a 'breaking' change?

Lipata:
@damyanpetev There is already an [issue](https://github.com/IgniteUI/help-topics/issues/128), to add this as a breaking change.

ig-ealbert:
I can make these changes next week in the docs.ja files, or we can wait until I do my next comparison of the files from the build (in less than 2 weeks).  That will cover these changes and any others that are made in the .js file comments.  What do you prefer @Lipata?  I don't think it's necessary to alert me of these things as long as we keep the current comparison process with one exception - I'll stop checking after code freeze, so changes after that should be assigned.

Lipata:
Thanks, I will remove the localization label

src/js/modules/infragistics.ui.editors.js
```diff
@@ -4055,7 +4055,7 @@
 				```
 				*/
 			maxDecimals: null,
```
damyanpetev: @Lipata Actually I see a set in the snippet - so this change should probably also address the behavior when setting this options with an existing value.

PR #783, opened 2017-01-30
#744 Decimals exceptions

### LINKED ISSUE #744: 'maxDecimals'/'minDecimals' behaviour should be consistent when set at initialization and at runtime
### Description  
* 'minDecimals' and 'maxDecimals' options can be set to a negative value at runtime. 
* It is possible to set 'minDecimals' value to an empty string
* 'maxDecimals' error message when trying to set its` value to an empty string should be improved

### Steps to reproduce  

1. Use the sample at http://jsfiddle.net/8j5wxywu/3/
2. Set the 'minDecimals' or 'maxDecimals' to a negative value (for example, `$("#numericEditor").igNumericEditor("option", "maxDecimals", -10)` )
3. Set the 'minDecimals' to an empty string - `$("#numericEditor").igNumericEditor("option", "minDecimals", "")`
4. Set the 'maxDecimals' to an empty string - `$("#numericEditor").igNumericEditor("option", "maxDecimals", "")`
 
### Result  
**After step 2:** 
Setting 'minDecimals' / 'maxDecimals' to a negative number is allowed
**After step 3:** 
Setting "minDecimals" to an empty string is allowed
**After step 4:**
The following error is shown: "Uncaught Error: Runtime changes are not allowed for the following option: maxDecimals"

### Expected result  
**After step 2:** 
Setting 'minDecimals' / 'maxDecimals' to a negative number is **not** allowed
**After step 3:** 
Setting "minDecimals" to an empty string is not allowed
**After step 4:**
The error message should say that 'minDecimals'/'maxDecimals' can be set to a number inside the '0 - 20' range only.

Lipata:
Closes #744 

Throw exceptions, when minDecimals/maxDecimals receive wrong values - initialization and runtime.


src/js/modules/infragistics.ui.editors.js
```diff
@@ -4543,15 +4572,18 @@
...
-						throw new Error($.ig.Editor.locale.setOptionError + option);
-					} else {
+						throw e;
+					}
+					if (this.options[ option ] !== prevValue) {
 						this._processInternalValueChanging(this.value());
 						if (!this._editMode) {
 							this._editorInput.val(this._getDisplayValue());
 						}
+						this._validateDecimalMinMax();
```
damyanpetev: This should probably be before _processInternalValueChanging :)

src/js/modules/infragistics.ui.editors.js
```diff
@@ -4460,22 +4461,31 @@
...
 
 			// A.M. October 11 2016 #420 "Spin button increase/decrease button not disabled"
 			if (this.options.buttonType === "spin") {
-				this._setSpinButtonsState(initialValue);
+				this._setSpinButtonsState(this.options.value);
 			}
 			if (this.options.spinDelta !== 1) {
 				delta = this.options.spinDelta;
 				if (typeof delta !== "number") {
-					this.options.spinDelta = 1;
```
damyanpetev: These should probably stay - at this stage the Error won't break the editor entirely, so spin still needs a delta 😕
(whether that should be the expected behavior is a different issue)

src/js/modules/infragistics.ui.editors.js
```diff
@@ -4499,11 +4509,30 @@
...
+				boundary = 15;
+			} else if (mode === "float") {
+				boundary = 7;
 			}
-			if (this.options.value < 0) {
-				this._editorInput.addClass(this.css.negative);
+
+			if (value === "" || isNaN(value) ||
+				(!isNaN(value) && (value < 0 || value > boundary))) {
+				throw new Error($.ig.util.stringFormat($.ig.Editor.locale.decimalNumber,
```
damyanpetev: Should this also set the valid boundary, same reason like spin delta above?
// cc @IG-Stanimir-Dimitrov

IG-Stanimir-Dimitrov: @damyanpetev  I will add a new issue to throw the exceptions at initialization time (i.e. break the editors) in case we have such configuration errors. 
//cc @kdinev

ig-ealbert:
You can use the following translation for Japanese.
I'll update our European file to remove the second entry.
dataMode が '{0}' の場合、{1} オプションは 0 ～ {2} の間の値を承諾します。

PR #795, opened 2017-02-01
#714 Add combo and editors 'disable' knockout binding handler

### LINKED ISSUE #714: Create Knockout binding for combo and editors that enables/disables them
Create binding handler, similar to the default KO ["disabled"](http://knockoutjs.com/documentation/disable-binding.html) handler, that will enables/disables the following controls - combo, editors.

The names of the bindings could be igComboDisable and igEditorEnable. We already have similar override for the default knockout "visible" binding. More about it can be found in the [following topic](http://www.igniteui.com/help/igcombo-knockoutjs-support).

The following configuration should work after the implementation - checking the checkbox should enable/disable the combo (the same is valid for the editors):

```
function ViewModel(actorsList) {
    this.isDisabled =  ko.observable(true);
}

<div id="combo" data-bind="igCombo: { width: '200' }, igComboDisable: isDisabled "></div>
<input type="checkbox" data-bind="checked: isDisabled"/>
```

Lipata:
Closes #714  .  

Add combo and editors 'disable' knockout binding handler.

### DIFF
--- a/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
+++ b/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
@@ -226,4 +226,15 @@
             combo.css("display", visible() ? "inline-block" : "none");
         }
     };
+
+    ko.bindingHandlers.igComboDisable = {
+        update: function (element, valueAccessor) {
+            var disabled = valueAccessor(),
+                combo = $(element);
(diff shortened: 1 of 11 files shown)


IvayloG:
There are several excess commented our references in the test files which can be removed.

PR #800, opened 2017-02-02
Fix for bug 231214: When you update a formatted field it can't be found by the allFields search filtering. Add QUnit tests

m-hristov:


src/js/modules/infragistics.datasource.js
```diff
@@ -5567,6 +5571,39 @@
 			}
 			return true;
 		},
+		_getFieldsWithFormatter: function () {
+			/* return array of fields(from the data schema) that have property "formatter" defined
+			if schema is not defined OR there aren't such fields - ruturns empty array */
+			if (!this.schema()) {
+				return [];
+			}
+			var i, f = this.schema().fields(), res = [];
```
kdinev: Move the `var` definition about the schema check and `return res;` if there is no schema.

PR #805, opened 2017-02-03
Filter by text implementation

mpavlinov:
Adding new API method: filterByText: function (expression, fields) 
/* Filters the data source locally by text. If "fields" parameter is set search is performed only in the listed fields otherwise all fields are searched.*/
filterByText is proxy to the filter method which was extended to include filtering by text. This is done so that all the filtering logic is not duplicated in both methods.


src/js/modules/infragistics.datasource.js
```diff
@@ -5493,6 +5497,146 @@
 
 			return fields;
 		},
+		_splitFilterExpression: function (search) {
```
kdinev: Each of the newly added methods is fully unit testable. However, I see that in the newly added unit tests, only the public method `filterByText` is invoked. Please include unit tests for all methods.

src/js/modules/infragistics.datasource.js
```diff
@@ -5493,6 +5497,146 @@
...
+			f = this.schema().fields();
+			for (i = 0; i < f.length ; i++) {
+				if (f[ i ].formatter) {
+					res.push(f[ i ]);
+				}
+			}
+			return res;
+		},
+		_generateFormattedRecords: function (data) {
+			var i, j, f, len, fr = [], schema = this.schema(),
```
kdinev: How will `filterByText` work when there is a checkbox column. The formatter will return the markup of the checkbox, so would the method filter by the markup string?

kdinev: Include this in the test cases as well.

skrustev: Since currently this is implementation on DataSource level and not on Grid level similar scenarios are not supported. 
If the users sets the field formatter though they will need to handle it themselves since we cannot predict what the user would use.

kdinev: @skrustev The thing is, if you have a checkbox field anywhere in the grid (a Boolean field, with `renderCheckboxes: true`) then the search will resolve these in this method. My question is, how would this affect the search results?

kdinev: @skrustev this cannot be included in the tests in the OSS version, so you should include it into the internal tests.

skrustev: For the checkbox specific case this shouldn't affect the results because this is handled by the rendered and the data is not changed whether the cells display checkboxes instead of values.

kdinev: Correct, but what the formatter returns is not the value in the date, is it?

PR #807, opened 2017-02-03
Fix for bug 231326. Page does not scroll when interacting with grid with no scrollbars.

skrustev:


MayaKirova:
I can still reproduce this on Win10 + touch on FireFox version 51.0.1. Touch interactions don't scroll the page if the grid has no scrollbars.
Note: touch events were explicitly enabled in the FireFox config- dom.w3c_touch_events.enabled set to 1. They are disabled by default.

MayaKirova:
Seems to be fixed in the newest version of FireFox - 52.0.

PR #818, opened 2017-02-08
Fix for issue #230963. Use the document boundaries if no proper posit…

dkamburov:
Fix for issue #230963

If the popover does not find proper position to be show up in the current window, use the document as a boundary and do another run finding proper position.


MayaKirova:
In 17.1 the popover always shows below the element. This might be due to some other changes specific to the 17.1 branch.

PR #819, opened 2017-02-09
igDataSource GroupBy Summaries implementation and tests

MayaKirova:
igDataSource GroupBy Summaries implementation and tests


MayaKirova:
Requested changes are applied with the latest commit.

PR #821, opened 2017-02-09
Fix for issue #820. Error is thrown when modifying prototypes.

### LINKED ISSUE #820: [igScroll] Error is thrown when the Array prototype is being modified and syncedElemsH/syncedElemsV options are being set.
### Description
Error is thrown due to use of for..in.. loop when the Array prototype is being modified for example without checking with hasOwnProperty.

### Steps to reproduce  
1. Run the attached sample

### Result  
The following error is thrown:
Uncaught TypeError: Cannot read property 'createDocumentFragment' of undefined

### Expected result  
There should be no errors

### Attachments  
[defaultTable-igScroll.gz](https://github.com/IgniteUI/ignite-ui/files/763653/defaultTable-igScroll.gz)

skrustev:
Resolves #820 .


src/js/modules/infragistics.ui.scroll.js
```diff
@@ -1197,7 +1197,7 @@
 		_linkElementsH: function (inElements) {
 			this._linkedHElems = [];
 			if (inElements) {
-				for (var index in inElements) {
+				for (var index = 0; index < inElements.length; index++) {
```
kdinev: Use the `hasOwnProperty()` check instead. Leave the for-in loop, but wrap everything inside in a hasOwnProperty check:

```
for (var index in inElements) {
    if (inElements.hasOwnProperty(index)) {
        ...
    }
}
```

That should fix your issue.

skrustev: Using the normal "for" loop also fixes the issue.
The "for..in" loop is usually used when going through properties of an object. In the cases I've used it normally it is expected to be an array so I think it more proper to use the normal for loop.

kdinev: @skrustev OK, if you expect an array then a regular for-loop works for this case.

PR #823, opened 2017-02-09
Fix for issue #820. Error is thrown when modifying prototypes. Merge 16.2

skrustev:
#820


src/js/modules/infragistics.ui.scroll.js
```diff
@@ -1142,7 +1142,7 @@
 		_linkElementsH: function (inElements) {
 			this._linkedHElems = [];
 			if (inElements) {
-				for (var index in inElements) {
+				for (var index = 0; index < inElements.length; index++) {
```
kdinev: Same comment as the PR in master.

PR #827, opened 2017-02-16
Added bottom border when virtualization is enabled

d-georgiev-91:


MayaKirova:
This has the same issues as the following pull request:
https://github.com/IgniteUI/ignite-ui/pull/828

Should be fixed in a few more themes as described in the other pull request.

PR #828, opened 2017-02-16
Added bottom border when virtualization is enabled

d-georgiev-91:


MayaKirova:
There doesn’t seem to be a border for the following themes:
-	Bootstrap -> only less file seems to be updated. The css file should also be updated.
-	Bootstrap4 -> the css and less file should be updated.

Also for the ios theme there seems to be too many borders:
ios

d-georgiev-91:
@MayaKirova iOS theme should be refactored

MayaKirova:
In that case a separate issue should be logged for the iOS theme.
I've verified that this is fixed for the other themes.

PR #831, opened 2017-02-17
Add new style classes regarding the GroupBy Summaries implementation.

skrustev:
Related to TFS: 232331.


kdinev:
@MayaKirova Could you leave a review as well after testing it?

PR #848, opened 2017-03-06
#777 Set numeric value to min/max if it's out of range on init

### LINKED ISSUE #777: [igEditors] minValue/maxValue options are not respected at initialization
### Description  
If we initialize an editor with value greater than the maxValue or with value below the minValue - minValue and maxValue options are not respected. 

### Steps to reproduce  

1. 
```
	value: 5,
        minValue: 9
});
```

### Result  
The value is 5.

### Expected result  
The value is 9

Note: This is reproduced for the igPercent editor also. 
This scenario is working fine in build 16.1.2020.
The issue is reproduced in build 16.1.2110.

damyanpetev:
Closes  #777  .  

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4479,6 +4479,15 @@
 				this._getRegionalOption("numericMinDecimals") :
 				this.options.minDecimals;
 		},
+		_setInitialValue: function (value) { // NumericEditor
+			// D.P. 6th Mar 2017 #777 'minValue/maxValue options are not respected at initialization'
+			if (!isNaN(this.options.maxValue) && value > this.options.maxValue) {
+				value = this.options.maxValue;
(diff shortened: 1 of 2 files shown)


kdinev:
@damyanpetev any ideas what happened to the coverage?

damyanpetev:
@kdinev The only file I did change went from 90.69 to 90.7 (which seems to be reflected correctly now that master also fell to 76%). Seems to be a coveralls update.
Look at the history from https://coveralls.io/builds/10398659 to https://coveralls.io/builds/10458790 - line coverage is about the same (~79%) but now there's also a "branches covered (71.55%)" metric weighing it down.
So the new math is (20708 + 14471) / (26169 + 20225) ~=75.827%.

kdinev:
@damyanpetev Thanks for looking into this. I noticed the coverage went down for all the PRs, so I figured there is something fishy with coveralls.

PR #866, opened 2017-03-14
Fix issue 809 Wrong value is set when we have 'isLimitedToListValues: true' and 'revertIfNotValid: false'

### LINKED ISSUE #809: [NumericEditor] Wrong value is set when we have 'isLimitedToListValues: true' and 'revertIfNotValid: false'
### Description  
Wrong value is set when changing the value when isLimitedToListValues: true, 	revertIfNotValid: false

### Steps to reproduce  

1. 
```
	buttonType: "dropdown",
	listItems: [3, 4, 5],
	isLimitedToListValues: true,
	revertIfNotValid: false
});
```
2. Click inside the editor
3. Spin up using the UP arrow key
4. Select and Delete the input value
5. Blur the editor

### Result  
The value is 1.7976931348623157e+308

### Expected result  
The value should be an empty string

IvayloG:
Closes #809  .  

Adding check for isLimitedToListValues and tests.


tests/unit/editors/Bugs/tests.html
```diff
@@ -1289,6 +1301,33 @@
...
+				input.val(null);
+				editor.blur();
+				equal(editor.igNumericEditor("value"), null, "Value should be null when allowNullValue=true");
+				editor.igNumericEditor("option","allowNullValue", "false");
+				editor.igNumericEditor("option","revertIfNotValid", "true");
+				editor.select();
+				keyInteraction(38, input); //Arrow Up
+				editor.blur();
+				editor.select();
+				keyInteraction(55, input); //7
```
damyanpetev: This should probably actually set the edit field value as it triggers the key events, otherwise there's no value to revert when the blur hits.

tests/unit/editors/Bugs/tests.html
```diff
@@ -1289,6 +1301,33 @@
...
+				input = editor.igNumericEditor("field"),
+				editor.select();
+				keyInteraction(38, input); //Arrow Up
+				input.val("");
+				editor.blur();
+				equal(editor.igNumericEditor("value"), "", "Value should be empty string");
+				editor.igNumericEditor("option","allowNullValue", "true");
+				editor.select();
+				keyInteraction(38, input); //Arrow Up
+				input.val(null);
```
damyanpetev: This will result in an empty string actually, if you want to test the null case you'll either need a different null value or use the value method instead.

tests/unit/editors/Bugs/tests.html
IvayloG: updated

tests/unit/editors/Bugs/tests.html
IvayloG: updated

kdinev:
@damyanpetev Are you approving after the changes, or you're still waiting for something additional?

damyanpetev:
@kdinev There are actually some branches after this change that might be unreachable (the coverage drop) and @IvayloG will be adding more tests and potentially adjustments as well.

PR #867, opened 2017-03-14
Adding component descriptions in loader.

petkozhekov:
Adding component descriptions in loader.


kdinev:
@petkozhekov There is a JSCS error. Can you take a look, because it says that a duplicate key is used?

Issue #876, opened 2017-03-16
Warning popup is displayed when maxValue date is selected from the dropdown calendar.


damyanpetev:
### Description  
Warning popup is displayed when maxValue date is selected from the dropdown calendar.

[16.1] [16.2] [TF233032]

### Steps to reproduce  

1. Run the attached sample 
2. Click the dropdown calendar of the editor.
3. Select September/08/2017 in the calendar.

### Result  
Warning popup is displayed saying value reached a max of "2017/09/08" and was set to that value.

### Expected result  
Since the selected is the max value defined, there should be no warning

### Attachments  
[sample.zip](https://github.com/IgniteUI/ignite-ui/files/846899/sample.zip)

damyanpetev:
The change for this is already in master from the changes for https://github.com/IgniteUI/ignite-ui/issues/536

Issue #536, opened 2016-11-18
Initial date object change


IG-Stanimir-Dimitrov:
Extends issue - Dates submitted through the igDateEditor/igDatePicker are always wrong #460 

When altering already created date objects, support submitting seconds/milliseconds as 00 (i.e. reset the initial values).

kdinev:
This would be a behavioral change and should be communicated with customers before performed. Currently the expectation of the issue is if the input format of the date does not contain, let's say seconds, the seconds to be set at 00. Correct me if I'm wrong! :)

IG-Stanimir-Dimitrov:
Correct!
To avoid breaking current FJ implementation, the editor will default to 00 the values that are not supplied only when instantiating new date object. When changing existing dates, nothing except the altered values will be changed.

PR #878, opened 2017-03-16
Adding test for #876

damyanpetev:
Adding just the test from PR #877 as the fix is already in applied in master.


kdinev:
@damyanpetev Should these tests be passing even though the functionality is in a different pull request?

damyanpetev:
@kdinev Yes the change that fixes #876 is already in master for #536. I just added the relevant bug test.

PR #910, opened 2017-03-29
fix issues in igLoader

hanastasov:
fix issues in igLoader


hanastasov:
Still doing work on this !.

PR #926, opened 2017-04-06
#896 spinWrapAround doesn't spin to minValue if there is no maxValue set

### LINKED ISSUE #896: [NumericEditor] spinWrapAround doesn't spin to minValue if there is no maxValue set
### Steps to reproduce  

```
$("#editor").igNumericEditor({
	dataMode: 'double',				
	minValue: 5,
	spinWrapAround : true,
	buttonType: 'spin'
});
```
2. Enter value 6.
3. Use the spinDown button and click it twice.
4. When the maximum number appears, click on the SpinUP button.
### Result  
The value doesn't change.

### Expected result  
The value should return to minValue.

dafo:
Closes #896


Marina-L-Stoyanova:
Use the following steps to reproduce the issue:
Enter value 6.
Use the spinDown button and click it twice.
When the maximum number appears, blur the editor and click on the SpinUP button.

The issue reproduces in this case.

dafo:
@Marina-L-Stoyanova, 
This behaviour is caused by a JS limitation - if the dataMode selected is 'float', 'long' or 'double' then the value in edit mode is represented in scientific notation but in blur mode is not. There isn't a non-hacky way to handle the comparison so it is best to update the documentation with an explanation about this scenario. 
If the user wants to achieve this scenario without setting the maxValue himself and she wants to use dataMode which maxValue is represented in scientific notation, she needs to set the 'scientificFormat' option for the editor.

Marina-L-Stoyanova:
@dafo Ok I will wait for you to make the changes in the help topic and then we can close this issue.

PR #936, opened 2017-04-12
Throw an exception when groupSeparator is equal to decimalSeparatorEd…

### LINKED ISSUE #852: When `decimalSeparator: ","` the decimal separator is lost after blurring the editor
### Description  
When the decimalSeparator is set to ',' and we have a fractional number, when we blur the editor the comma sign disappears making the number an integer.

### Steps to reproduce  

1. 
	decimalSeparator: ",",
	value: 1.1
});
2. Focus the editor
3. Blur the editor

### Result  
The value is 11

### Expected result  
The value is 1,1

### Attachments  

[sample.zip](https://github.com/IgniteUI/ignite-ui/files/827214/sample.zip)

dafo:
Closes #852  .

### DIFF
--- a/src/js/modules/i18n/infragistics.ui.editors-en.js
+++ b/src/js/modules/i18n/infragistics.ui.editors-en.js
@@ -74,7 +74,8 @@
 				centuryThresholdValidValues: "centuryThreshold property shoud be between 0 and 99.",
 				noListItemsNoButton: "No spin or dropdown button is rendered because there are no listitems.",
 				decimalNumber: "When dataMode is '{0}', the {1} option can accept numeric values between 0 and {2}.",
-				decimalSeparatorErrorMsg: "The decimalSeparator option requires a different value. Its value should be a single character."
+				decimalSeparatorErrorMsg: "The decimalSeparator option requires a different value. Its value should be a single character.",
+				decimalSeparatorEqualsGroupSeparatorErrorMsg: "The groupSeparator option requires a different value than the decimalSeparator."
 			}
 		};
 	}
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4081,6 +4081,7 @@
 			/* type="string" Gets/Sets the character, which is used as decimal separator.
 				Note: this option has priority over possible regional settings.
 				Note: Even if the default value is null - if internationalization file is provided and it contains default values for those properties the values are imlicitly set.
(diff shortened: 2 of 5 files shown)


src/js/modules/i18n/infragistics.ui.editors-en.js
```diff
@@ -74,7 +74,8 @@
 				centuryThresholdValidValues: "centuryThreshold property shoud be between 0 and 99.",
 				noListItemsNoButton: "No spin or dropdown button is rendered because there are no listitems.",
 				decimalNumber: "When dataMode is '{0}', the {1} option can accept numeric values between 0 and {2}.",
-				decimalSeparatorErrorMsg: "The decimalSeparator option requires a different value. Its value should be a single character."
+				decimalSeparatorErrorMsg: "The decimalSeparator option requires a different value. Its value should be a single character.",
+				decimalSeparatorEqualsGroupSeparatorErrorMsg: "The groupSeparator option requires a different value than the decimalSeparator."
```
Lipata: Please add the English message to the other localization files.

bazal4o: Let's make the message something like: The groupSeparator and the decimalSeparator options can not have equal values.

ig-ealbert:
I translated the content for Japanese.  It is pending review.  I added the content to our outsourcing file for 17.1 for other languages.

ig-ealbert:
Can we use the following translations for Japanese?  I could add them through pull request once they're in master, or they can be added by the author here.  We'll outsource the European and Russian languages around code freeze, assuming all content will be added by then.
decimalSeparatorErrorMsg: "decimalSeparator オプションには別の値が必要です。値を単一文字に設定してください。",
decimalSeparatorEqualsGroupSeparatorErrorMsg: "groupSeparator および decimalSeparator オプションを同じ値に設定できません。"

dafo:
I have added the translations for Japanese but I`ve also found that the following string is missing in the Japanese file: setOptionError. I`ve added this but can we  also translate this to Japanese? I will update it once you provide me the string.
Or maybe just the name of the string is changed:
cannotSetRuntime: "このオプションはランタイムに設定できません。",
 setOptionError: 'Runtime changes are not allowed for the following option: ',

If the translation is the same, I will just replace the name of the string.

ig-ealbert:
@dafo Yes, the name seems to have changed.  If you could replace the 。 at the end with a : , that would be great.  Otherwise it's ok as is.  Thank you for looking into it.

dafo:
This is done.

PR #943, opened 2017-04-13
#942 Fix and adding tests.

### LINKED ISSUE #942: When clearing with the 'clear' button, the value is set to 0 even if 0 is not in the list of items
### Description  
When we clear the value using the clear button, the value is set to 0 even if zero is not in the list of items.

### Steps to reproduce  

1. 
`$('#editor').igNumericEditor({
	buttonType: "dropdown, clear",
	listItems: [3, 4, 5],
	isLimitedToListValues: true,
	revertIfNotValid: false
});`
2. Enter edit mode and select '3'
3. Blur the editor
4. Click the 'clear' button

### Result  
Value is 0

### Expected result  
Value is empty string.

IvayloG:
When isLimitedToListItems -->Setting "" instead of 0 when isLimitedToListValues.

Closes #942  .


IvayloG:
*Make sure to check after #809 and #942 are merged to test the following scenario - reverting to a previously existing value:
1. Use the sample from #942.
2.set isLimitedToListValues: true
3. Check if correctly reverts to a previously existing value. If a value, is selected from the list, blur, then focus the input and enter invalid value (outside the list of values.)
-check if the value is reverted when blur 
-check if the value is reverted when clearing via the clear button

kdinev:
@IvayloG There is a conflict now :(

PR #963, opened 2017-04-24
Closes #962: Remote GroupBy is not working with remote Paging (Known Issue) 

### LINKED ISSUE #962: Remote GroupBy is not working with remote Paging (Known Issue)
### Description  
Remote GroupBy is not working with remote Paging(Known Issue - from 16.2+). 
In ig.DataSource paging is applied ONLY for visible row(including data records and metadata rows-groupby rows). Thus remote paging and remote grouping are not supported. It should be able paging to be applied ONLY for data records and non-data rows(like metadata rows - group-by rows and summary group-by rows) should not be included in paging processing.

### Steps to reproduce  

1. Open the attached sample
2. Observe consolo log.

### Result  
When grouping is applied - all rows(data records and group-by rows) are included in page processing

### Expected result  
It should be able to apply paging ONLY for data records. Metadata records - as group-by and summary group-by - should NOT be included in page processing.

### Sample
[sample(2).zip](https://github.com/IgniteUI/ignite-ui/files/955072/sample.2.zip)

m-hristov:
Closes #962

### DIFF
--- a/src/js/modules/infragistics.datasource.js
+++ b/src/js/modules/infragistics.datasource.js
@@ -4372,8 +4372,8 @@
 				/*A.T. workaround for jQuery's 1.5 and above bug related to dataFilter and success callback.
 				We need to explicitly set the dataType to "text" when manually parsing it */
 				/* get jquery version */
-				if (jQuery.fn.jquery) {
-					ver = jQuery.fn.jquery.split(".");
+				if ($.fn.jquery) {
+					ver = $.fn.jquery.split(".");
 				}
 				if (ver && ver.length >= 2) {
 					/* if jQuery is 1.5 and greater or if the first major version is greater than 1 (when jQuery 2 comes out)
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.datasource.js
```diff
@@ -4359,8 +4372,8 @@
 				/*A.T. workaround for jQuery's 1.5 and above bug related to dataFilter and success callback.
 				We need to explicitly set the dataType to "text" when manually parsing it */
 				/* get jquery version */
-				if ($.fn.jquery) {
-					ver = $.fn.jquery.split(".");
+				if (jQuery.fn.jquery) {
```
kdinev: @skrustev Didn't we make some changes to always use `$` instead of mix of `$` and `jQuery` because of webpack, or is it the other way around?

m-hristov: Yep, it should be used $ instead of jQuery. I didn't get latest version and I made a pull request without latest changes that @skrustev did. I just committed with the appropriate use of $ :)

PR #974, opened 2017-05-02
Add new date patterns for regional files, needed by the igScheduler.

Lipata:


ig-ealbert:
@Lipata Could you please explain what's going on here?  I thought that the content of these regional files was taken from jQuery or jQuery UI.  Is that still the case?  The only one I can easily review is the Japanese one, which doesn't make sense with "yy" (like "17"?).  What's the use case and is this how jQuery does it?

Lipata:
@ig-ealbert we need two new formats for the igScheduler: April 2017, 20 April 2017. That's why they have been added to the regional files. The format is different from the others, cause we are using jQuery formatter, which is different from the formats using in the igEditors. We will make them consistent for the next release. The two formats in Japanese will produce the following dates: 2017年4月 and 2017年4月15日. If you think this is ok for Japanese, you can approve the change.
For the patterns of the other languages I have used the existing patterns and have created the patterns for the igScheduler. For those two patterns it is important the order of the day, month and year. Other symbols are not included.

ig-ealbert:
@Lipata What is "yy" in jQuery?  In .NET I think it's the last two digits of the year, so like "17".  If that's the case, then wouldn't these not result in the correct output?
dateTitleFullPattern: 'yy年m月d日',
dateTitleMonthPattern: 'yy年m月',
Does "yy" in jQuery resolve to the full year with 4 digits?  Then what's the difference with "yyyy"?
If the results are what you write (2017年4月 and 2017年4月15日), then they're good, and I can approve this.  Please let me know about "yy" and then I'll do the review.

src/js/modules/i18n/regional/infragistics.ui.regional-pl.js
```diff
@@ -43,6 +43,8 @@
 	    dateTimePattern: 'yyyy-MM-dd HH:mm',
 	    timePattern: 'HH:mm',
 	    timeLongPattern: 'HH:mm:ss',
+		dateTitleFullPattern: 'dd MM yy',
```
Lipata: @IG-Stanimir-Dimitrov thank you! It's fixed.

Lipata:
@ig-ealbert In a sample, where you have jQuery UI script loaded, you can try in the console the following code:
$.datepicker.formatDate("yy年m月d日", new Date()) and see the result.
As I said, now those two new patterns are different from the ones for the igEditors, and this will be consistent for the next release. After all the results are the expected ones.

PR #975, opened 2017-05-03
#810 fix colors for selected and the active list items

### LINKED ISSUE #810: The selected and the active list items have the same styles applied
### Description  
The item that is selected has the same color as the one that is just active. It would be good to distinguish those states.

### Steps to reproduce  

1.
```
	buttonType: "dropdown",
	listItems: [1, 4, 5]
});
```
2. Open the dropdown list
3. Click '1'
4. Open the list again
5. Press the DOWN arrow key

### Result  
Botsh items '1' and '4' have the same look

### Expected result  
The color of the active item shouldn`t be the same as the selected one.

### Attachments  
screenshot

desig9stein:
Closes #810 .  

#810 fix colors for selected and the active list items and add some styles for infragistics.igniteui.theme.less. 
update the gitignore file to ignore all files .idea/


src/css/themes/bootstrap/LESS/infragistics.theme.css
```diff
@@ -1011,6 +1011,10 @@ body .ui-tooltip {
   font-weight: normal;
   border: 0;
 }
+/* Bug #741 S.D. 20.04.2017 When bootstrap theme used with bootstrap css text in editor can't be seen when editor has focus*/
```
desig9stein: Svilen was committed the change only in the .less and now I commit the generated CSS file

src/css/themes/flatly/infragistics.theme.css
```diff
@@ -2550,7 +2552,7 @@ body .ui-tooltip {
 .ui-iggrid .ui-iggrid-mergedcellsbottom.ui-state-highlight {
   border-bottom-color: #ecf0f1;
 }
```
desig9stein: Again the change was committed in .less and now I commit the generated CSS file.

src/css/structure/modules/infragistics.ui.splitter.css
```diff
@@ -25,6 +25,12 @@
 	width: 100%;
 }
 
+/* Bug 236424 - 04.05.2017 Marin Popov - Horizontal scrollbar is displayed... */
```
desig9stein: This is the fix for  BUG-236424 which is in tfs

damyanpetev:
I've verified the fix for #810 - the editor list items match interaction styles with the combo as desired and are distinct in the main Infragistics theme in all others, except the old 2012 one which uses the same color for everything on the combo, but at least hover and selection are different on the editor. The last part seems completely intentional, so I think we are good.

PR #977, opened 2017-05-03
#955 Fixing problems when spinDelta larger than 1.

### LINKED ISSUE #955: DateEditor with spin delta allows spinning to invalid hours
### Description  
DateEditor with spin delta allows spinning to invalid hours

### Steps to reproduce  

1. Run the sample from the attachment
2. Click on the hours and make sur that cursor is showing
3. Press down arrow

### Result  
The editor value becomes -2:00

### Expected result  
User should ne be able to spin to invalid hours

It is possible to spin up to 53:00

### Attachments  

[igDateEditor-spin-delta.zip](https://github.com/IgniteUI/ignite-ui/files/943651/igDateEditor-spin-delta.zip)

Lipata:
Closes #955 .


damyanpetev:
@Lipata After further testing, this is not quite there yet - when you spin hours in the AM/PM format after you jump up from 11 PM to 1 AM (setup used in the test) and you keep spinning up until 11AM the next spin jumps back to 1AM instead of 1PM. So you can only spin withing the morning hours, only if you spin back down would it go to 11PM and still it gets stuck in PM hours as well. Something in the transition is not right.

damyanpetev:
@Lipata Check out the last test case (should fail) going up to "03 12:00 AM", when it spins down it goes to 7AM instead of 7PM.
Also when there's no date part on the mask, but we're still using "date" data mode (2nd case) and the hours keep spinning but the day doesn't (unlike case 3).
Tagging @IG-Stanimir-Dimitrov for those.

IG-Stanimir-Dimitrov:
@damyanpetev, I would expect the date to spin up/down together with the hours when limitSpinToCurrentField is false. If users want to restrict the spinning just to hours, then limitSpinToCurrentField must be set to true

damyanpetev:
@IG-Stanimir-Dimitrov I agree, it's just the combination of hours-only mask and full Date mode is not really a meaningful combination - it's a bit unexpected for the developer to care of the date part or the user to be aware of it. `limitSpinToCurrentField` doesn't really clearly define the expected result in that case either, but again it's not a common use anyway.
In any case, the current logic spinning while editing is completely separate and based on the value available in the mask so this result it to be expected, but we might change that in the future.

Lipata:
When spin delta is in less than the maximum value of current time period, e.g. for 12 hour format, the spin should be < 12, then the editor will work. There is a [separate issue](https://github.com/IgniteUI/ignite-ui/issues/978) for the case when value is > 12.

PR #984, opened 2017-05-03
Fix util files code throwing errors when using Webpack with the combined files.

skrustev:


gmurray81:
Sorry, not sure I get notifications for updates to these for some reason, need to check my notification settings.

PR #987, opened 2017-05-05
[igCombo] The editor text changes even if selectionChanging event is canceled when allowCustomValue is set to true 

### LINKED ISSUE #986: [igCombo] The editor text changes even if selectionChanging event is canceled when allowCustomValue is set to true
### Description  
Canceling the `selectionChanging` event still triggers a change in the combo input when `allowCustomValue` is set to true.

### Steps to reproduce  

1. Run the attached sample
2. Enter some text which does not exist in the data source of the combo.
3. Select "test1" from the dropdown list.

### Result  
The "test1" is not selected in the dropdown list (as expected) but the text in the combo input is changed.

### Expected result  
The text in the combo input should not change.

### Attachments  
[igCombo-sample.zip](https://github.com/IgniteUI/ignite-ui/files/978893/igCombo-sample.zip)

rkaraivanov:
Closes #986


src/js/modules/infragistics.ui.combo.js
```diff
@@ -6784,6 +6784,7 @@
                 }
 
                 if (noCancel) {
+                    options.selCanceled = false;
```
kdinev: Don't add properties to the `options` object. Use `_options` instead.

kdinev:
I'm not setting this to `status: awaiting-test` yet because the fix will be reworked.

PR #988, opened 2017-05-05
[igCombo] The editor text changes even if selectionChanging event is canceled when allowCustomValue is set to true 16.2

rkaraivanov:
Closes #987


src/js/modules/infragistics.ui.combo.js
```diff
@@ -6782,6 +6782,7 @@
                 }
 
                 if (noCancel) {
+                    options.selCanceled = false;
```
kdinev: Don't add properties to the `options` object. Use `_options` instead.

kdinev:
I'm not setting this to `status: awaiting-test` yet because the fix will be reworked.

PR #991, opened 2017-05-11
Num scientific fixes

### LINKED ISSUE #761: Value is not an empty string no matter that the editor is empty when setting the value to a number with too many digits
### Description  
When we set the value to a number with too many digits after the decimal point, an exception is thrown but the value of the editor is set to that value no matter that the editor is empty.

### Steps to reproduce  

1. Initialize a numeric editor with the following options:
```
dataMode: 'double',
value: 0.000000000000000000001
```
2. Take a look at the Console
3. Get the 'value' of the editor

### Result  
The following error is shown: 
Uncaught RangeError: toFixed() digits argument must be between 0 and 20

The value method returns 1e-21

### Expected result  
What is the expected result after following the steps to reproduce?

The value method returns an empty string

damyanpetev:
Closes #761  .  

After looking at the root cause of the error I've updated most of the handling for numbers in scientific format:
- Handle decimal separator correctly when parsing and displaying/editing
- Fixed `scientificFormat` when applying formatting sign (not just E casing)
- Respect `maxDecimals` and `roundDecimals` on applicable modes and numbers, but also handle correctly values that can't fit fixed-point representation
- Removed some redundant code


dafo:
1. Run the numeric editor with the following option set:
`scientificFormat: "e+"`
2. Paste "23e+1" inside the editor
3. Blur

Actual result: 
2.3e+2
Value is 230

Expected result:
2.3e+1
Value is 23

damyanpetev:
@dafo if you paste "23e+1" it's somewhat expected to be normalized to 2.3e+2 and the end value should be 230, but not in edit mode right after you paste - which was not being formatted by the paste/insert handler. Should be good now.

PR #993, opened 2017-05-12
adding css for marker colors

ig-robstoffers:
Added support for the chart markers to pull default marker brushes and outlines from css so add the css to the Infragistics theme.

Closes # .  


kdinev:
@ig-robstoffers I believe those changes should be made to the .LESS file, instead of in the CSS file.

ig-robstoffers:
Yep I will do that instead.  Thanks!

PR #995, opened 2017-05-12
Properly revert display value in mask editor.

### LINKED ISSUE #903: [MaskEditor] EmptyChars are not respected when we call value method.
### Description  
PadChars are not respected when we change the initial value, but leave the required positions empty.

### Steps to reproduce  

```
$("#maskEditor").igMaskEditor({
	inputMask: "CCC00",
	padChar: "+",
	emptyChar: "/",
       value: "aaa"
});
```
2. Enter new value like: "sss"
3. Blur the editor

### Result  
The value is "aaa//"

### Expected result  
The value should be "aaa++"

Lipata:
Closes #903


IvayloG:
As this is a functionality change, it will be required to update the API documentation regarding the padChar option as well and remove the following section:
 "Note that this option is visible, only when the **revertIfNotValid** option is set to **false**."

PR #1028, opened 2017-05-19
Fix the problem when there is a .touch class on mobile and desktop

desig9stein:


mpavlinov:
The problem still reproduces for 17.1

PR #1029, opened 2017-05-19
Fix the problem when there is a .touch class on mobile and desktop

desig9stein:


kdinev:
This change is already in master

PR #1044, opened 2017-06-15
I18n

mpavlinov:
Removing localization files dependencies.


kdinev:
@mpavlinov are those not needed?

PR #1047, opened 2017-06-16
Hover background matches text color 16.2 in superhero theme

desig9stein:
Feature chooser buttons hover background matches text color in superhero theme


kdinev:
@desig9stein This fix should also be PRed to 17.1, correct?

desig9stein:
@kdinev Yes I did the PR for 17,1

PR #1048, opened 2017-06-16
Fix hover background matches text color

desig9stein:
Feature chooser buttons hover background matches text color


kdinev:
@desig9stein I noticed the changes in this PR are a bit different from the one for 16.2. There is colors changed in all bootstrap themes in this one.

desig9stein:
@kdinev It's different because of the difference between the file structure, this is normal.

PR #1052, opened 2017-06-21
Runtime locale changes - initial implementation

### LINKED ISSUE #1049: Editors render redundant attributes (title vs. aria-label) on buttons
### Description  
The Editor controls render redundant attributes on the drop down buttons, spin buttons, and clear buttons. Both `title` and `aria-label` are rendered for the elements with `role=button`.

### Steps to reproduce  

1. Check either of these methods:
```
_renderDropDownButton
_renderSpinButtons
_renderClearButton
```

### Result  
The result is that under some screen readers all attributes will be read to the user, resulting in text of the sort:
"Button, Spin Up, Spin Up Button"

### Expected result  
Redundant attributes should be avoided in order to improve the accessibility of the control. The control is accessible now, but it may be a little _overly accessible_ at times.

### Attachments  
Attach a sample if available, and screenshots, if applicable.

kdinev:
Closes #1049


build/packages/combined-files.js
```diff
@@ -113,6 +113,7 @@ module.exports = {
                     "./dist/js/modules/infragistics.util.jquerydeferred.js",
                     "./dist/js/modules/infragistics.datasource.js",
                     "./dist/js/modules/infragistics.templating.js",
+					"./dist/js/modules/infragistics.ui.widget.js",
```
ChronosSF: Looks like a small indentation issue.

src/js/infragistics.loader.js
```diff
@@ -874,12 +882,13 @@ $.ig.dependencies = [
 	{
 		widget: "igCombo",
 		dependency: [
-			{ name: "igUtil" },
+			{ name: "igWidget" },
+			{ name: "igWidget" },
```
ChronosSF: The line is duplicated.

src/js/modules/infragistics.ui.combo.js
```diff
@@ -32,6 +32,7 @@
 			"jquery-ui",
 			"./infragistics.util",
```
ChronosSF: If widget depends on util do we need to specify the combo depends on both util and widget?

src/js/modules/infragistics.ui.combo.js
```diff
@@ -1620,6 +1621,9 @@
             };
 
             this._analyzeOptions();
+
+			// Options related to localization
+			this._changeLocale();
```
ChronosSF: Looks like a small indentation issue. Could be because the other lines use spaces.

src/js/modules/infragistics.ui.widget.js
```diff
@@ -0,0 +1,119 @@
+/*!@license
+ * Infragistics.Web.ClientUI Toolbar <build_number>
```
ChronosSF: Maybe a copy-pasted string "Toolbar"?

src/js/modules/infragistics.ui.widget.js
```diff
@@ -0,0 +1,119 @@
...
+                });
+
+                // Get
+                var regional = $(".selector").igWidget("option", "regional");
+
+                // Set
+                $(".selector").igWidget("option", "regional", "ja");
+            ```
+            */
+            regional: "en-US"
```
ChronosSF: Looks to be using spaces - the review page has indentation issues.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5176,14 +5173,14 @@
...
-						$.ig.util.stringFormat($.ig.Editor.locale.maxValExceedSetErrMsg,
+						$.ig.util.stringFormat(this._getLocaleValue("maxValExceedSetErrMsg"),
 							this.options.maxValue));
 				} else if (!isNaN(this.options.minValue) && newValue < this.options.minValue) {
 					newValue = this.options.minValue;
 
 					// Raise Warning level 2
 					this._sendNotification("warning",
-						$.ig.util.stringFormat($.ig.Editor.locale.minValExceedSetErrMsg,
+						$.ig.util.stringFormat(this._getLocaleValue("maxValExceedSetErrMsg"),
```
ChronosSF: Please, check if you didn't change minValExceedSetErrMsg with maxValExceedSetErrMsg

src/js/modules/infragistics.ui.editors.js
```diff
@@ -5758,7 +5755,7 @@
 
 							// Raise Warning level 2
 							this._sendNotification("warning",
-								$.ig.util.stringFormat($.ig.Editor.locale.minValExceedSetErrMsg,
+								$.ig.util.stringFormat(this._getLocaleValue("maxValExceedSetErrMsg"),
```
ChronosSF: Please, check if you didn't change minValExceedSetErrMsg with maxValExceedSetErrMsg

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6071,15 +6068,15 @@
...
+						$.ig.util.stringFormat(this._getLocaleValue("maxValExceedSetErrMsg"),
 							this.options.maxValue));
 				} else if (!isNaN(this.options.minValue) &&
 					newValue / this.options.displayFactor < this.options.minValue) {
 					newValue = this.options.minValue * this.options.displayFactor;
 
 					//Notify
 					this._sendNotification("warning",
-						$.ig.util.stringFormat($.ig.Editor.locale.minValExceedSetErrMsg,
+						$.ig.util.stringFormat(this._getLocaleValue("maxValExceedSetErrMsg"),
```
ChronosSF: Please, check if you didn't change minValExceedSetErrMsg with maxValExceedSetErrMsg

src/js/modules/infragistics.ui.combo.js
kdinev: Nope we don't. I'll fix this.

kdinev: I'll also remove jquery and jquery-ui, since they are dependencies for widget as well.

src/js/modules/infragistics.ui.widget.js
kdinev: Not maybe. Certainly.

src/js/modules/infragistics.ui.widget.js
```diff
@@ -0,0 +1,137 @@
...
+
+                // Set
+                $(".selector").igWidget("option", "regional", "ja");
+            ```
+            */
+			regional: "en-US"
+		},
+		_setOption: function (option, value) {
+			var prevValue = this.options[ option ];
+			if (prevValue === value) {
```
ChronosSF: As discussed, if locale and regional can be an object, getting the option value, changing some props and setting it back will not result in a change because of this line.

kdinev:
@ChronosSF Addressed.

kdinev:
@m-hristov I checked the tile manager code and the reference to the splitter locale is not a correct implementation and should be changed. The error really belongs to the tile manager context, because the option is not settable in the tile manager context. This will be changes to reference an error message from the tile manager locale.

I don't think that implementation which includes passing a widget name as a parameter in each function, and then calling getWidgetName again in each function is good.

src/js/modules/infragistics.ui.widget.js
```diff
@@ -0,0 +1,133 @@
...
+    /*
+		Base igWidget for all Ignite UI widgets.
+	*/
+    $.widget("ui.igWidget", {
+		localeWidgetName: null,
+        options: {
+            /* type="string|object" Set/Get the locale setting for the widget.
+            ```
+                 //Initialize
+                $(".selector").igWidget({
```
bazal4o: If we want the same code snippet to appear in all the inherited widget we should use %%WidgetName%% placeholder instead of igWidget

kdinev:
@m-hristov The structure of the null texts should follow the locale format.

PR #1053, opened 2017-06-21
Add some ES6 modularized versions of util

gmurray81:
Note:
This probably shouldn't be merged as is. The intent is to break up util into some ES6 modules, that can be imported individually when using bundling tools. However, we will still want to have the combined version of infragistics.util.js for use in simpler scenarios or when using a dynamic module loader, etc. I propose we use these files as the system of record, but then create some grunt tasks to combine them to form the equivalent of the current infragistics.util.js? Does that sound reasonable? For that file to become build output? The combination process would need to remove the ES6 imports and exports, and throw the AMD IFFE back around all the content, I think.


kdinev:
Including @damyanpetev to this conversation.

kdinev:
@damyanpetev Reminder that you need to look at this PR.

damyanpetev:
Sorry for the delay.

While I like the idea in general, I do have some questions:
- This affects just the infragistics.util.js file and the jquery ones remain untouched, right? I.e. keeping the current breakdown for how the final product is to be consumed?
-  If I get this correctly the idea is to break down the util file just to make it more manageable for us. So besides that and adding a build step to combine it back into the usual single file (with the AMD IIFE in the template).. do we also ship the extra module files as well? Guessing no seeing and there's no dependency breakdown for those parts into other modules and no other use for those besides being combined back.
- If so, what's the point of the ES imports again?
- Last but not least, I think there's a bunch of code that's really there for the sake of it. Case in point, the kinda ridiculous $.ig.util.sleep that I can't find being used anywhere. Likely not the only one.

If the breakdown is indeed intended internal only, I say we give it a go. At the very least it'll make cleaning up that code easier.

gmurray81:
> This affects just the infragistics.util.js file and the jquery ones remain untouched, right? I.e. keeping the current breakdown for how the final product is to be consumed?

the idea is that infragistics.util.js should look as current in terms of how it is included in the final product, but will now be produced by combining these modular files.

> If I get this correctly the idea is to break down the util file just to make it more manageable for us. So besides that and adding a build step to combine it back into the usual single file (with the AMD IIFE in the template).. do we also ship the extra module files as well? Guessing no seeing and there's no dependency breakdown for those parts into other modules and no other use for those besides being combined back.

We don't necessarily ship those files. This is more to enable us to use those files in other contexts other than the Ignite product (at least, not the current version). In other scenarios, say, angular, we will want to import these util modules individually. I can show you, in context, how some of our DV angular prototypes do this, to further understanding. I think offline would be best for that though.

> If so, what's the point of the ES imports again?

Angular, and also consumption of Ignite using bundling tools potentially. But I can't go into much detail on that here. The present Ignite product would just use the combined result, though.

> Last but not least, I think there's a bunch of code that's really there for the sake of it. Case in point, the kinda ridiculous $.ig.util.sleep that I can't find being used anywhere. Likely not the only one.

Some of that code is used by the translated output, although probably not that sleep guy, hopefully ....

gmurray81:
I'm going to close this PR, I think perhaps this is superceded by the in process TS stuff.

gmurray81:
Also, I think I need to purge my fork, it isn't playing nice.

PR #1062, opened 2017-06-26
Adding fix for task 236973 - Checkboxes cannot be clicked

desig9stein:

### DIFF
--- a/src/css/structure/modules/infragistics.ui.tree.css
+++ b/src/css/structure/modules/infragistics.ui.tree.css
@@ -135,8 +135,8 @@
 }
 .ui-igtree .ui-igcheckbox-normal-off,
 .ui-igtree .ui-igcheckbox-normal-on {
-	width:100%!important;
-	height:100%!important;
+	width:100%;
+	height:100%;
 }
 /* Bug 236973 END */
 
--- a/src/css/structure/modules/infragistics.ui.tree.css
+++ b/src/css/structure/modules/infragistics.ui.tree.css
@@ -125,6 +125,21 @@
 	height:1.6em;
 }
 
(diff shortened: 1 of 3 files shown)


src/css/structure/modules/infragistics.ui.tree.css
```diff
@@ -125,6 +125,21 @@
...
+.ui-igtree  .ui-igcheckbox-normal-on {
+	position:relative;
+	top:0;
+	left:0;
+	margin-top:0;
+	margin-left:0;
+}
+.ui-igtree .ui-igcheckbox-normal-off,
+.ui-igtree .ui-igcheckbox-normal-on {
+	width:100%!important;
```
kdinev: @desig9stein Is the !important really necessary here?

desig9stein: @kdinev I need this because of another important rule which is located in the Infragistics theme

/* icon SIZING FIXES for controls which expect a specific icon size */
.ui-iggrid tr .ui-icon,
.ui-igtree .ui-icon,
.ui-igedit .ui-icon,
.ui-igsplitter .ui-icon,
.ui-igcombo .ui-icon {
  height: 16px !important;
  width: 16px !important;
  font-size: 11px;
}

I can remove the row with:.ui-igtree .ui-icon, but I don't know if that will not break anything else

kdinev: @desig9stein Can you experiment with removing the !important in this general definition. Let's see if that would break the font icons, and if not then we should get rid of it :)

desig9stein: @kdinev Hi, I did some testing and so far I couldn't find anything that stops me from removing these !important rules, so I did it.

PR #1064, opened 2017-06-26
Update ui.igWidget

m-hristov:
Update _changeLocale, use one attribute - data-locale instead of data-localeid and data-localeattr. Improve _getLocaleValue to use $.ig.locale as well if options.locale is set to string


src/js/modules/infragistics.ui.widget.js
```diff
@@ -98,36 +99,93 @@
...
+				.find("[" + this.localeAttributeName + "]")
+				.each(function () {
+					self._changeLocaleForElement($(this));
+				});
+		},
+		_getLocaleOptions: function () {
+			return {};
+		},
+		_changeLocale: function () {
+			var opts = this._getLocaleOptions();
```
kdinev: I don't quite understand how this works. The `_setOption` for `locale` calls `_changeLocale`, which calls `_getLocaleOptions`, which returns an empty object `{}`, so `_changeLocale` is always a noop. How do you get to the rest of the methods?

src/js/modules/infragistics.ui.widget.js
```diff
@@ -98,36 +99,93 @@
...
+				}
+			}
 		},
-		_changeLocaleByKey: function (element, key, attr) {
-			if (attr) {
-				element.attr(attr, this._getLocaleValue(key));
-			} else {
-				element.text(this._getLocaleValue(key));
+		_changeLocaleByLocaleId: function (localeId, $container) {
+			if (!localeId) {
```
kdinev: What is the difference between `_changeLocaleById` and `_changeLocaleInContainer`? Both receive a `$container` parameter, but I don't see one calling the other. In fact, in `changeLocaleById` if `localeid` is not provided but `$container` is specified, the `_changeLocaleInContainer` is not called and `_changeLocaleById` becomes a noop. What are the use cases requiring two methods?

src/js/modules/infragistics.ui.widget.js
```diff
@@ -98,36 +99,93 @@
 				return $.ig.regional.defaults[ key ];
 			}
 		},
+		_getLocaleWidgetName: function () {
+			this.localeWidgetName = this.localeWidgetName ||
```
kdinev: What are the cases in which there isn't a `widgetFullName` and there is a `widgetName` and vice-versa?

src/js/modules/infragistics.ui.widget.js
```diff
@@ -98,36 +99,93 @@
...
+							(this.widgetFullName || this.widgetName)
+								.replace(/^((ui-ig(\.)*)|(ig(\.)*))/ig, "");
+			return this.localeWidgetName;
+		},
+		_getLocaleObject: function (locale) {
+			return $.ig.locale &&
```
kdinev: This method does not allow extracting a value from the locale option provided to the specific instance of the control.

src/js/modules/infragistics.ui.widget.js
```diff
@@ -98,36 +99,93 @@
...
-			var locale = this.options.locale,
-				widgetName = this.localeWidgetName || this.widgetName.split("ig")[ 1 ];
-
-			if (locale && locale[ key ]) {
-				return locale[ key ];
-			} else if ($.ig && $.ig[ widgetName ] && $.ig[ widgetName ].locale) {
-				return $.ig[ widgetName ].locale[ key ];
-			} else {
-				return "";
+			var locale = this.options.locale, localeObject, localeWidgetName;
```
kdinev: Why do you need a `_getLocaleObject` method, if further locale object processing is done in other methods calling it?

src/js/modules/infragistics.ui.widget.js
m-hristov: In case of this.options.locale is string then we should get which are specific locale options and upate widget with new values (I have used widget this._setOptions)
Otherwise if this.options.locale is object probably we should just call this._setOptions(this.options.locale);

src/js/modules/infragistics.ui.widget.js
m-hristov: _changeLocaleInContainer - changes all elements that have attribute data-locale.
_changeLocaleById - changes all elements with the specified localeId.
When _setOption is called for specific locale option then call _changeLocaleById.

src/js/modules/infragistics.ui.widget.js
m-hristov: AFAIK it is recommended to try using widgetFullName(in older jQuery versions widgetName was only available)

src/js/modules/infragistics.ui.widget.js
kdinev: In both cases the method is a `$.noop`.

src/js/modules/infragistics.ui.widget.js
kdinev: So does this mean all widgets have to handle locale changes in the widget in order to decide which one they should be calling?

src/js/modules/infragistics.ui.widget.js
m-hristov: Yep, but the idea was if locale is string - my version of this widget is implemented for the case when this.options.locale is string. I have in mind that it could be an object so I could easily change it.

src/js/modules/infragistics.ui.widget.js
m-hristov: They are helper function and should be called ONLY when there is need to be called by the widgets(usually in _setOption).

PR #1066, opened 2017-07-03
Implement issue #1060. Make custom/native scrollbars more flexible in general.

### LINKED ISSUE #1060: [igScroll] Allow single scroll bar to take full width/height when available
[igGrid_scrollbars.zip](https://github.com/IgniteUI/ignite-ui/files/1122605/igGrid_scrollbars.zip)
### Description  
The current scroll bars keep a static right/bottom space regardless of there being a second bar. It would be more visually consistent if the bars can span the full heights/width dynamically if there's just one scroll direction.

### Steps to reproduce  

1. Open https://www.igniteui.com/scroll/configuration-options
2. Reduce columns to 3-4 so only vertical scroll remains
3. Observe the space beneath it

### Result  
The space for the no-longer needed horizontal scroll is still kept

### Expected result  
Would be nice if the bar would switch to the full height instead.

### Attachments  
image

skrustev:
Not using anymore fixed widths/heights. Add through CSS classes depending on if a single custom/native scrollbar is visible, it shouldn't display empty area.

Resolves #1060

### DIFF
--- a/src/css/structure/modules/infragistics.ui.scroll.css
+++ b/src/css/structure/modules/infragistics.ui.scroll.css
@@ -42,10 +42,17 @@
 	position: absolute;
 	right: 0px;
 	top: 0px;
+	height: 100%;
 }
 
 .igscroll-vtrack {
+	top: 15px;
+	bottom: 30px;
+	position: absolute;
+}
 
+.igscroll-vtrack-singlescrollbar {
+	bottom: 15px !important;
 }
 
 /* Vertical arrow - general */
@@ -93,6 +100,20 @@
 }
 
 /* Arrow Down */
+.igscroll-downarrow {
+	bottom: 15px;
+	position: absolute;
+}
+
+.igscroll-downarrow-singlescrollbar {
+	bottom: 0px !important;
+}
+
+.igscroll-downarrow-active {
+	bottom: 15px;
+	position: absolute;
+}
+
 .igscroll-downarrow:before {
 	-webkit-transform: rotate(-180deg);
 	-moz-transform: rotate(-180deg);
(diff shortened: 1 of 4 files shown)


src/css/structure/modules/infragistics.ui.scroll.css
```diff
@@ -93,6 +100,20 @@
 }
 
 /* Arrow Down */
+.igscroll-downarrow {
+	bottom: 15px;
+	position: absolute;
+}
+
+.igscroll-downarrow-singlescrollbar {
+	bottom: 0px !important;
```
kdinev: Is `!important` really necessary?

src/css/structure/modules/infragistics.ui.scroll.css
```diff
@@ -146,11 +167,19 @@
...
 .igscroll-htrack {
 	height: 15px;
 	float: left;
+	left: 15px;
+	right: 30px;
+	position: absolute;
+}
+
+.igscroll-htrack-singlescrollbar {
+	right: 15px !important;
```
kdinev: Same about `!important`

src/css/structure/modules/infragistics.ui.scroll.css
```diff
@@ -199,6 +228,20 @@
 }
 
 /* Arrow Right */
+.igscroll-rightarrow {
+	right: 15px;
+	position: absolute;
+}
+
+.igscroll-rightarrow-singlescrollbar {
+	right: 0px !important;
```
kdinev: Same.

src/css/structure/modules/infragistics.ui.scroll.css
skrustev: I have some more commits that address this and the native scrollbars as well, so this shouldn't be merged yet.

PR #1078, opened 2017-07-14
#1074 Check if an drop down item is defined as observable

### LINKED ISSUE #1074: KO apply bindings is very slow with igCombo
### Description  
In "ko.bindingHandlers.igComboList = { ... " where ko.applyBindingsToNode is executed, a slight improvement can be implemented in order to reduce the loading time. Add isObservable check in the for loop.

### Steps to reproduce  

1. Open the attached sample

### Result  
Loading is taking more than 40 seconds

### Expected result  
Application to be loaded immediately 

### Attachments  

[ALotOfDataKO.zip](https://github.com/IgniteUI/ignite-ui/files/1128036/ALotOfDataKO.zip)

Lipata:
Check if an drop down item is defined as observable and only then apply binding.


tests/unit/knockout/combo/tests.html
```diff
@@ -78,9 +78,36 @@
 			{ name: "name3333", value: "value3" }
 		];
 
+		modelBigData = [
```
damyanpetev: This could be generated, right? :)

tests/unit/knockout/combo/tests.html
```diff
@@ -211,214 +240,222 @@
...
-			$chk.remove();
-		});
+			var testId = 'Test big data';
+			test(testId, function () {
+				var $combo = $("#bigDataContainer");
+				$combo.append($("<span id='comboBigData'></span>").attr("data-bind", "igCombo: { dataSource: bigData, textKey: 'name', valueKey: 'id', width: '200', mode: 'dropdown', enableClearButton: false }"));
+				var start = (new Date()).getTime();
+				ko.applyBindings(globalViewModel, $combo[0]);
+				var end = (new Date()).getTime();
+				ok(end - start < 5000, "Binding big data is taking too much time");
```
damyanpetev: Not sure this is the best way to go about this. Would prefer to have the test fail at the timeout, rather than waiting

tests/unit/knockout/combo/tests.html
Lipata: Done!

tests/unit/knockout/combo/tests.html
Lipata: The problem is that ko.applyBindings, which is too slow, is not a promise, so using a timeout will not help.

PR #1088, opened 2017-07-21
Implement zoombar rutime locale change

dkamburov:


kdinev:
@dkamburov Update the loader as well please.

PR #1091, opened 2017-07-27
Runtime locale

kacheshmarova:
Closes #ISSUE_NUMBER


src/js/modules/i18n/infragistics.datasource-de.js
```diff
@@ -9,22 +9,19 @@
...
-			"jquery"
-		], factory );
+		define( [], factory );
 	} else {
-		factory(jQuery);
+		factory();
 	}
 }
-(function ($) {
+(function () {
```
mpavlinov: Add $ = $ || {}; for all of the files

Issue #1042, opened 2017-06-14
When you set value larger than the maxLength, the input is cleared.


Marina-L-Stoyanova:
### Description  
In igTextEditor if you have set maxLength property and then set a value larger than the maxLength through the value method/value option, the value gets cleared.

### Steps to reproduce  

1. Create a sample with the following code:

```
$("#editor").igTextEditor({
	placeHolder: "Test",
	maxLength : 5
 });
```

2. Enter a value that is shorter than 5 characters. 
3. Change the value through the value method: `$("#editor").igTextEditor("value","abcdefg")`

### Result  
The value in the editor is cleared. The notification says that the value is trimmed to 5 symbols.

### Expected result  
The value should not be cleared it should be trimmed.

>Note: When you verify this issue make sure you test this scenario too: [https://jsfiddle.net/hhyecpbh/](https://jsfiddle.net/hhyecpbh/)

Issue #481, opened 2016-11-03
[igTextEditor]Set listItems option runtime


agicheva:
**Steps to reproduce**
Create a code sample with the following code:
```
$(".selector").igTextEditor();
 
$(".selector").igTextEditor("option", "listItems", ["item 1", "item 2", "item 3"]);
```

**Actual result**
Throws an error in the console.

**Expected result**
It should throw a proper message.

**Comment**
The option has to be described in the API documentation.

PR #1094, opened 2017-07-27
Trim value, when its length is larger then the maxLenght one (17.1).

Lipata:
#1042

(cherry picked from commit 5834588f26ac9826b72810bdd352d4b1bdfe470c)

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -1079,9 +1079,17 @@
 					this._updateValue(newValue);
 					this._editorInput.val(this._getDisplayValue());
 				} else {
-					this._clearValue();
-					if (this._focused !== true) {
-						this._exitEditMode();
+
+					// N.A. July 27th, 2017, #1042: Trim value, when its length is larger then the maxLenght one.
+					if (this.options.maxLength) {
+						newValue = newValue.toString().substring(0, this.options.maxLength);
+						this._updateValue(newValue);
+						this._editorInput.val(this._getDisplayValue());
+					} else {
+						this._clearValue();
+						if (this._focused !== true) {
+							this._exitEditMode();
+						}
 					}
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -1079,9 +1079,17 @@
 					this._updateValue(newValue);
 					this._editorInput.val(this._getDisplayValue());
 				} else {
-					this._clearValue();
-					if (this._focused !== true) {
-						this._exitEditMode();
+
+					// N.A. July 27th, 2017, #1042: Trim value, when its length is larger then the maxLenght one.
+					if (this.options.maxLength) {
+						newValue = newValue.toString().substring(0, this.options.maxLength);
```
bazal4o: substring is zerobased - so you should trim to this,options.maxLength -1

src/js/modules/infragistics.ui.editors.js
```diff
@@ -1079,9 +1079,17 @@
 					this._updateValue(newValue);
 					this._editorInput.val(this._getDisplayValue());
```
bazal4o: change this line with something like:
 if (this._focused !== true) { 
   this._exitEditMode(); 
}

src/js/modules/infragistics.ui.editors.js
```diff
@@ -1079,9 +1079,17 @@
...
 				} else {
-					this._clearValue();
-					if (this._focused !== true) {
-						this._exitEditMode();
+
+					// N.A. July 27th, 2017, #1042: Trim value, when its length is larger then the maxLenght one.
+					if (this.options.maxLength) {
+						newValue = newValue.toString().substring(0, this.options.maxLength);
+						this._updateValue(newValue);
+						this._editorInput.val(this._getDisplayValue());
```
bazal4o: change this line with something like:
 if (this._focused !== true) { 
   this._exitEditMode(); 
}

Lipata: Having in mind that, when editor can be updated through API method - value, insert, it will also require updating the visual editor value (no matter if it is focused or not), so I suggest leaving it that way.

tests/unit/editors/textEditor/tests.html
```diff
@@ -1575,6 +1575,18 @@
 			ok($('#maxLengthTest').parents().hasClass("ui-ignotify-warn"), "The notifier is not shown");
 			$editor.remove();
 		});
+
```

PR #1095, opened 2017-07-27
Trim value, when its length is larger then the maxLenght one.

### LINKED ISSUE #1042: When you set value larger than the maxLength, the input is cleared.

Lipata:
Closes #1042

(cherry picked from commit e70c9f2892bb6b404d9bbe4b9a9bedb734cac6bc)

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -1079,9 +1079,17 @@
 					this._updateValue(newValue);
 					this._editorInput.val(this._getDisplayValue());
 				} else {
-					this._clearValue();
-					if (this._focused !== true) {
-						this._exitEditMode();
+
+					// N.A. July 27th, 2017, #1042: Trim value, when its length is larger then the maxLenght one.
+					if (this.options.maxLength) {
+						newValue = newValue.toString().substring(0, this.options.maxLength);
+						this._updateValue(newValue);
+						this._editorInput.val(this._getDisplayValue());
+					} else {
+						this._clearValue();
+						if (this._focused !== true) {
+							this._exitEditMode();
+						}
 					}
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -1079,9 +1079,17 @@
...
 				} else {
-					this._clearValue();
-					if (this._focused !== true) {
-						this._exitEditMode();
+
+					// N.A. July 27th, 2017, #1042: Trim value, when its length is larger then the maxLenght one.
+					if (this.options.maxLength) {
+						newValue = newValue.toString().substring(0, this.options.maxLength);
+						this._updateValue(newValue);
+						this._editorInput.val(this._getDisplayValue());
```
bazal4o: This should be applied according to focused state.

Lipata: Having in mind that, when editor can be updated through API method - value, insert, it will also require updating the visual editor value (no matter if it is focused or not), so I suggest leaving it that way.

Lipata:
#1103 was merged instead of this, due to a wrong commit after merge.

tests/unit/editors/textEditor/tests.html
```diff
@@ -1576,6 +1576,18 @@
 			ok($('#maxLengthTest').parents().hasClass("ui-ignotify-warn"), "The notifier is not shown");
 			$editor.remove();
 		});
+
```

Issue #892, opened 2017-03-22
[TextEditor] Value is null even when 'nullValue: 0'


Marina-L-Stoyanova:
### Description  
null is returned even when 'nullValue' is set to 0.

### Steps to reproduce  
1. Initialize editor like this one:
```
$('#editor').igTextEditor({
        allowNullValue: true,
	nullValue: 0
})
```
2. Try calling the value method.
`$("#editor").igTextEditor("value")`
### Result  
value is null

### Expected result  
value is 0

**NOTE** Issue #779 was the same but for the Numeric Editor.

PR #1099, opened 2017-07-28
Dpetev/mask initial error

### LINKED ISSUE #1027: Exception is thrown if "nullValue" is set to a number in the MaskEditor
### Description  
"Uncaught TypeError: Cannot read property 'length' of undefined" is thrown if you try to type inside a mask editor that has "nullValue" option's value set to a number.

### Steps to reproduce  

1. `$("#editor").igMaskEditor({ allowNullValue: true, nullValue: "0" });`
2. Click inside the editor
3. Type "abc"

### Result  
Exceptions are thrown:
infragistics.ui.editors.js:26   Uncaught TypeError: Cannot read property 'length' of undefined

### Expected result  
The editor is rendered with "0_____" dispalyed inside.
No exceptions are thrown when typing

damyanpetev:
Closes #1027 

Since the original issue was in incorrect initialization, I've made sure _setInitialvalue is always called.
This also obsoletes and reverts most changes made for #892.
The change actually led to some additional changes for text and numeric editors that did not rely on that being called and for some other edge cases.

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -1804,14 +1804,6 @@
 			switch (option) {
 			case "value":
 				this.value(value);
-
-				//if (this._validateValue(value)) {
-				//	this._updateValue(value);
-				//	this._exitEditMode();
-				//} else {
-				//	this._clearValue();
-				//	this._exitEditMode();
-				//}
 				break;
 			case "placeHolder":
 				this._applyPlaceHolder();
@@ -6940,7 +6932,6 @@
 			return dataModeValue;
 		},
 		_updateValue: function (value) { //igMaskEditor
-			var nullValue;
 			if (value === "") {
 
 				// Convert empty value by dataMode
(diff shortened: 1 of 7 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -4478,6 +4486,11 @@
 				this.options.minDecimals;
 		},
 		_setInitialValue: function (value) { // NumericEditor
+			if (typeof value === "undefined") {
```
bazal4o: Double check if this adds a breaking change.

damyanpetev: Indeed I was wondering if this should default to nothing or 0 - the unmodified behavior with wrong `nullValue` on the numeric would be empty string, no errors on typing, but unable to clear value which defaults to 0 instead. Will update the initial to stay empty as test definitions match that, but have a leftover `value: ""`.. 😞 
I was wondering if this should default to nothing or 0

PR #1101, opened 2017-07-31
Implement getRegionalValues in util and refactor $.ig.formatter function

dkamburov:

### DIFF
--- a/src/js/modules/infragistics.util.jquery.js
+++ b/src/js/modules/infragistics.util.jquery.js
@@ -122,7 +122,7 @@
 	//checkbox markup classes can be changed
 	//and setting them in a a box can be done by the predefined classes "ui-state-default ui-corner-all ui-igcheckbox-small"
 	$.ig.checkboxMarkupClasses = "";
-	$.ig.getRegionalOption = function (reg) {
+	$.ig.getRegionalOptions = function (reg) {
 		if (!reg) {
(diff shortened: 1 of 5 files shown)


src/js/modules/infragistics.util.jquery.js
```diff
@@ -122,236 +122,277 @@
 	//checkbox markup classes can be changed
 	//and setting them in a a box can be done by the predefined classes "ui-state-default ui-corner-all ui-igcheckbox-small"
 	$.ig.checkboxMarkupClasses = "";
+	$.ig.getRegionalOptions = function (reg) {
```
kdinev: Don't these two belong to the base widget, same as `getLocaleValue`, or only `getRegionalValue`?

dkamburov: The util is not a widget, so I cannot extend the base widget here.

kdinev: My question is why `getRegionalValue` belongs to util. It's strange to me that there is a function which takes the object it extracts from as a parameter and is on a utility level.

Can you show the examples of how, and from where this method is used, apart from the formatter? Also, how is this integrated in the editors?

dkamburov: Let's have a meeting to discuss this.

PR #1103, opened 2017-08-01
#1042 Trim value, when its length is larger then the maxLenght one.

### LINKED ISSUE #1042: When you set value larger than the maxLength, the input is cleared.

Lipata:
Closes #1042
Closes #1095

### DIFF
--- a/tests/unit/editors/textEditor/tests.html
+++ b/tests/unit/editors/textEditor/tests.html
@@ -1578,14 +1578,15 @@
 		});
 		
 		testId = "Setting listItems on run time";
-		test(testId, 2, function () {
+		test(testId, 3, function () {
 			var editor = $('#issue481'), items = ["item 1", "item 2", "item 3"],
 			currentVal = editor.igTextEditor("option", "listItems");
 
 			ok(currentVal === null, "ListItems should not be set on initialization");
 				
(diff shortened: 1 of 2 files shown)


tests/unit/editors/textEditor/tests.html
```diff
@@ -1576,6 +1576,18 @@
 			ok($('#maxLengthTest').parents().hasClass("ui-ignotify-warn"), "The notifier is not shown");
 			$editor.remove();
 		});
+
```

Lipata:
This test is not for that issue. This test was for #481 , but was not merged to 17.1 and 17.2. The test for this issue is pull request #1110

PR #1107, opened 2017-08-02
#1082 Add suppressKeyboard option for igDatePicker.

### LINKED ISSUE #1082: [igDatePicker] Add option for keyboard showing/hiding on touch devices
Provide a configuration option `suppressKeyboard` to suppress on-screen keyboard from showing when the date picker dropdown button is clicked.

Lipata:
Closes #1082

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -1007,10 +1007,6 @@
 			}
 		},
 		_setFocus: function (event) {
-			if (this._shouldNotFocusInput) {
-				event.target.blur();
-				return;
-			}
 
 			// D.P. 22nd Aug 2016 #226 Can't right-click paste in Edge, double focus event on menu closing
 			if (this._focused) {
@@ -11010,6 +11006,13 @@
 				this._spinDownButton.addClass(this.css.disabled);
 				this._detachButtonsEvents(this._spinDownButton);
 			}
+		},
+		_setFocus: function (event) {
+			if (this._shouldNotFocusInput) {
+				event.target.blur();
+					return;
+			}
+			this._super(event);
 		},
 		_setBlur: function (event) { // igDatePicker
 			if (this._pickerOpen) {
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -10843,7 +10843,22 @@
 			/* @Ignored@ This option is inherited from a parent widget and it's not applicable for igDatePicker */
 			listItems: null,
 			/* @Ignored@ This option is inherited from a parent widget and it's not applicable for igDatePicker */
-			listWidth: 0
+			listWidth: 0,
+			/* type="bool" Suppress on-screen keyboard from showing when the date picker dropdown button is clicked.
+				```
+				//Initialize
+				$(".selector").%%WidgetName%%({
+					suppressKeyboard : true
+				});
+
+				//Get
+				var readOnly = $(".selector").%%WidgetName%%("option", "suppressKeyboard");
+
+				//Set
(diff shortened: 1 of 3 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -1007,6 +1007,11 @@
 			}
 		},
 		_setFocus: function (event) {
+			if (this._shouldNotFocusInput) {
```
damyanpetev: If this flag applies only to the date picker it probably shouldn't be in the base `_setFocus`

Lipata: Done

src/js/modules/infragistics.ui.editors.js
```diff
@@ -1592,7 +1593,22 @@
 				$(".selector").%%WidgetName%%("option", "suppressNotifications", true);
 			```
 			*/
-			suppressNotifications: false
+			suppressNotifications: false,
+			/* type="bool" Suppress on-screen keyboard from showing when the dropdown button is clicked and the list is opened.
```
damyanpetev: If the option name is set for consistency, can we at least make this extra clear in the description?
- Option "prevents initial **or** removes focus when the drop button is clicked". Yes the use case is for mobile on-screen ones, but the actual functionality is generic. Also to avoid confusion that this is not a "mobile-readonly"
- There should be a `Note:` clearly stating that this option does no auto-detection whatsoever, meaning it's left to the application to figure out dynamically when to enable this.

//CC @IG-Stanimir-Dimitrov

src/js/modules/infragistics.ui.editors.js
```diff
@@ -11351,6 +11409,10 @@
 				if (currentInputValue) {
 					this._editorInput.val(currentInputValue);
 				}
+			} finally {
+				if (this.options.suppressKeyboard) {
```
damyanpetev: I think you can safely call delete here even without the check - just to make sure the flag always gets removed (even if option is changes by some unexpected handling)

src/js/modules/infragistics.ui.editors.js
IG-Stanimir-Dimitrov: @damyanpetev , we can go with something like for the description:

```
type="bool" Gets/Sets whether the onscreen keyboard should be shown when the dropdown button is clicked (touch devices only). This option prevents initial or removes focus when the drop button is clicked.

Note: The keyboard will still show when the input is focused in editable mode.
```

damyanpetev: @IG-Stanimir-Dimitrov I don't get the `(touch devices only)` bit though - that was my point for the note - there's no mobile device detection built in and I wanted that to be very clear.
The option could run just as well on non-touch desktops if someone prefers the "open dropdown without focus" behavior.
```
type="bool" Gets/Sets whether the onscreen keyboard (on touch devices) should be shown when the dropdown button is clicked . This option prevents initial focus or removes it when the drop button is clicked.

Note: The option does not perform device detection so its behavior is always active if enabled.
```
?
the "clicked" wording is also not super correct

dafo:
1. Tap the calendar button to show the calendar.
2. Tap 'previous month' or 'next month' button

Result: The soft keyboard is shown

Expected result: The soft keyboard is not shown

src/js/modules/infragistics.ui.editors.js
```diff
@@ -11124,6 +11128,11 @@
 					}
 				};
 			}
+			options.onChangeMonthYear  = function () {
```
damyanpetev: @Lipata If we handle jQuery datepicker events internally (like `onSelect`) those should save a possible such handler passed in with `datepickerOptions` by the user so those are properly called as well

PR #1110, opened 2017-08-03
#1042 Add forgotten test for maxLength option.

Lipata:
#1042


Lipata:
There are two problems here. The test in pull request #1094 was not for that issue.  It was for #481 . It had been forgotten to be merged to 17.1 and 17.2 at the first place. Therefore it was merged with commit #1094, but for some reason the test for fix in #1094 was neglected. This pull request add that missing test.

PR #1113, opened 2017-08-03
#1082 Add suppressKeyboard option for igDatePicker.

Lipata:
#1082


dafo:
1. Tap the calendar button to show the calendar.
2. Tap 'previous month' or 'next month' button

Result: The soft keyboard is shown

Expected result: The soft keyboard is not shown

PR #1115, opened 2017-08-03
Date.toStringFormat token replacement

davidnegley:
working on some time axis stuff for DataChart, task #239473.  the Date.toStringFormat() function is incomplete.  i added a few formats to the bottom of the method.  also, i'm using token replacement instead of trying to replace the entire format string.

note that the function no longer throws an exception, but instead returns the format string with whatever replacements were made.


kdinev:
@davidnegley Could you please include unit tests about this in util?

PR #1116, opened 2017-08-04
Implement runtime regional settings change

dkamburov:

### DIFF
--- a/src/js/modules/infragistics.ui.combo.js
+++ b/src/js/modules/infragistics.ui.combo.js
@@ -1965,9 +1965,31 @@
 			if ($.ig && $.ig.formatter) {
 				if (this.options.format === "auto" &&
 					($.type(item) === "date" || $.type(item) === "number")) {
-					item = $.ig.formatter(item, null, null);
+					item = $.ig.formatter(
+						item,
+						null,
+						null,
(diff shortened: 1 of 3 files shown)


src/js/modules/infragistics.ui.combo.js
```diff
@@ -1965,9 +1965,31 @@
 			if ($.ig && $.ig.formatter) {
 				if (this.options.format === "auto" &&
 					($.type(item) === "date" || $.type(item) === "number")) {
-					item = $.ig.formatter(item, null, null);
+					item = $.ig.formatter(
```
kdinev: Will the formatter be refactored to take an object of a type? Like `FormatterArgs`?

dkamburov: This seems like the better option, but we'll introduce a breaking change and we'll also have to change the [formatter topic](https://www.igniteui.com/help/formatting-dates-numbers-and-strings). If we are good with them, I'll refactor it and change the topic.

mpavlinov: It won't be breaking change if we test for the type of the first parameter and split the logic in the formatter based on that check.

PR #1121, opened 2017-08-08
#793 Move decimal options validation at the start of initialization so it …

### LINKED ISSUE #793: When minDecimals/maxDecimals are set with values outside the valid boundaries, the editors must not function at all
### Description  
When the igEditors minDecimals/maxDecimals options receive wrong values, the editors must not accept them and NOT function at all (applied to both initialization & runtime).

Relates to https://github.com/IgniteUI/ignite-ui/pull/783#pullrequestreview-19337545

damyanpetev:
…would fail with wrong values.

Closes #793

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4460,6 +4460,9 @@
 		},
 		_initialize: function () {
 			this._super();
+
+			// D.P. 8th Aug 2017 #793 Should not initialize with wrong decimal values.
+			this._validateDecimalSettings();
 			this._applyRegionalSettings();
 			this._applyDataModeSettings();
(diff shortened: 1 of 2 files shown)


Marina-L-Stoyanova:
@damyanpetev  I have noticed that in the test you use numericMaxDecimals and numericMinDecimals, the options should be maxDecimals and minDecimals.

damyanpetev:
@Marina-L-Stoyanova That's an older test I updated the because it ran with swapped min/max. The actual regional setting is actually `numericMaxDecimals` (since there's a separate one for percent and currency) and `_getRegionalValue` (now in `igWidget`) will also read it from the options. So this is pretty much the same as setting `maxDecimals` directly.

PR #1125, opened 2017-08-11
Fix for bug 238125 - When headerText contains HTML string the column cell data is broken (contains escaped html)

petkozhekov:
Fix for bug 238125 - When headerText contains HTML string the column cell data is broken (contains escaped html)

The issue is caused by the aria-label attribute which accepts the value of the headerText option. When there are html tags in the header text some of the symbols need to be escaped so the generated html is correct.

### DIFF
--- a/src/js/modules/infragistics.util.js
+++ b/src/js/modules/infragistics.util.js
@@ -399,7 +399,7 @@
 	//and setting them in a a box can be done by the predefined classes "ui-state-default ui-corner-all ui-igcheckbox-small"
 	$.ig.checkboxMarkupClasses = "";
 
-	$.ig.endcode = function (value) {
+	$.ig.encode = function (value) {
 		/* Encode string.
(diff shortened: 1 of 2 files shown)


kdinev:
There is an encode function in until. You should use it instead.

petkozhekov:
Currently the "encode" method is in the infragistics.templating.js file. If I use it in the infragistics.util.js file it will create a dependency between the two. Is it a good idea to move the encode method into the util file and then use it?

kdinev:
@petkozhekov At least create an encode function in util. I will think about adding a dependency in the templating file on util, and removing the encode method from the templating file. Btw, doesn't the `igGrid` already depend on templating, and thus it has access to encode coming from templating?

ChronosSF:
@kdinev The AMD support for templating already loads util and util.jquery .

PR #1128, opened 2017-08-11
#1102 Fix case where Validator null min/max throw an error

### LINKED ISSUE #1102: Setting valueRange min/max to null causes exceptions
### Description  
When setting the min or max values of the valueRange property to null, an unhandled exception will occur with certain values.

### Steps to reproduce  

1. Run [attached sample](https://github.com/IgniteUI/ignite-ui/files/1188675/page.zip).
2. In the top editor, type "5" and blur the element

### Result  
A "Cannot read property 'toString' of null" exception occurs.

### Expected result  
No exceptions should occur and the value should be valid.

### Additional Notes
The same exception occurs if you enter a negative number in the bottom editor.

The following logic in the isValid method of $.ig.igValidatorValueRule does not work well when nulls are used.

```
hasMin = typeof minValue === "number" || minValue,
hasMax = typeof maxValue === "number" || maxValue;

if ((hasMin || hasMax)) {
	if (isNumber && !options.date) {
		value = this._parseNumber(value, options);
		min = hasMin && minValue;
		min = value < min ? min.toString() : null;
		max = hasMax && maxValue;
		max = value > max ? max.toString() : null;
	}
```

Lipata:
Closes #1102 

Also refactoring the code a bit so it's actually readable..

### DIFF
--- a/src/js/modules/infragistics.ui.validator.js
+++ b/src/js/modules/infragistics.ui.validator.js
@@ -2272,10 +2272,12 @@ $.ig.igValidatorValueRule = $.ig.igValidatorValueRule || $.ig.igValidatorNumberR
 			if ((hasMin || hasMax)) {
 				if (isNumber && !options.date) {
 					value = this._parseNumber(value, options);
-					min = hasMin && minValue;
-					min = value < min ? min.toString() : null;
-					max = hasMax && maxValue;
-					max = value > max ? max.toString() : null;
+					if (hasMin) {
+						min = value < minValue ? minValue.toString() : null;
+					}
+					if (hasMax) {
+						max = value > maxValue ? maxValue.toString() : null;
+					}
 				} else if (isDateParsable && !options.number) {
 					value = new Date(value);
 					if (hasMin) {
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.validator.js
```diff
@@ -2257,58 +2257,61 @@ $.ig.igValidatorValueRule = $.ig.igValidatorValueRule || $.ig.igValidatorNumberR
...
-						maxValue = maxValue.toLocaleString();
-					}
-					max = value > max ? max.toLocaleString() : null;
-				}
+		if (!isDateParsable && !isNumber) {
+			//can't be handled by this rule
+			return true;
+		}
+		minValue = options.valueRange.push ? options.valueRange[ 0 ] : options.valueRange.min;
+		maxValue = options.valueRange.push ? options.valueRange[ 1 ] : options.valueRange.max;
```
mpavlinov: options.valueRange[ 1 ] will throw an unhandled error if there is only one element in the array. Although do API docs says that it should be 2 elements in the array I don't see an explicit check for that in the code.

damyanpetev: Actually it won't - it'll be undefined just as it would be if the option is set with just `{ min: 3}` object. There are tests running with just one item in the array too.

PR #1133, opened 2017-08-15
#978 Implement spinDelta as object

### LINKED ISSUE #978: Define igDateEditor/igDatePicker spinDelta option as a set of deltas for the different time periods
### Description  
In igDateEditor and igDatePicker spinDelta option should allow defining a set of different values, each of them applied to the different time periods (minutes, hours, days). We may want the spinDelta for the minutes to be 30, while for the hours to be 1. Now the spinDelta option allows only one value that is applied to all the time periods.

Lipata:
Implement spinDelta option of igDateEditor and igDatePicker to support object with individual deltas for each time period.

Closes #978

When control is initialized and the value of the spinDelta isn't correct, then exception is thrown. With this commit setting the value to 1 in that case is removed, cause control stops working when the error is thrown. 
The case is not the same when initially we have correct value for the spinDelta, but in run-time we try to set wrong one. In that case we also throw an exception, which can be caught by the application, but we revert to the previous correct value of the spinDelta. In that case we can have a working control, after the attempt to set wrong value, cause the control itself was working at the first point.


src/js/modules/i18n/infragistics.ui.editors-bg.js
```diff
@@ -60,6 +60,9 @@
 		btnValueNotSupported: 'Необходима е различна стойност на бутона. Изберете една от следните стойности: "dropdown", "clear" или "spin".',
 		scientificFormatErrMsg: 'Необходим е различен scientificFormat. Изберете една от следните стойности: "E", "e", "E+" или "e+".',
 		spinDeltaIsOfTypeNumber: "Необходим е различен тип на spinDelta. Трябва да бъде въведено положително число.",
+		spinDeltaIsOfTypeNumberForPeriod: "A different type of spinDelta is required for the {0}. A positive number between {1} and {2} should be entered.",
+		spinDeltaIsOfTypeNumberOrObject: "A different type of spinDelta is required. A positive number or an object, defining different time periods deltas, should be entered.",
```
damyanpetev: "different time period~s~ deltas"

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8080,6 +8094,44 @@
...
+					throw new Error(this._getLocaleValue("spinDeltaCouldntBeNegative"));
+				} else {
+					this.options.spinDelta = parseInt(delta, 10);
+				}
+			} else if (typeof delta === "object") {
+				deltaRanges = [ [ "years", 10 ], [ "months", 12 ], [ "days", 28 ],
+					[ "hours", this._dateIndices.hh24 ? 24 : 12 ], [ "minutes", 60 ],
+					[ "seconds", 60 ], [ "milliseconds", 1000 ] ];
+
+				for (index = 0; index < deltaRanges.length; index++) {
```
bazal4o: It's better to iterate through the option "delta" instead of delta ranges

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8131,7 +8183,23 @@
 			this._spin(this.options.spinDelta, true);
 		},
 		_handleSpinDownEvent: function () { // DateEditor
-			this._spin(-this.options.spinDelta, true);
+			this._spin(this._revertDelta(this.options.spinDelta), true);
+		},
+		_revertDelta: function (delta) {
+			var revDelta = { };
+			if (typeof delta === "object") {
+				$.each(delta, function (key, value) {
```
bazal4o: Check if valuable to use for key in delta instead of $.each (performance).

src/js/modules/infragistics.ui.editors.js
Lipata: The case is that if spinDelta for years is not defined (it's not in the options), then we need to set it to 1. That's why it's proper to iterate through deltaRanges array. The validation happens on initialization or when set dynamically at run-time, that's why it cannot cause performance issues.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7787,21 +7787,37 @@
 				spin type="string" Spin buttons are located on the right side of the editor
 			*/
 			buttonType: "none",
-			/* type="number" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
+			/* type="number|object" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
 				When not editing (focused) the delta is applied on the day if available in the input mask or the lowest available period.
 				When in edit mode the time period, where the cursor is positioned, is incremented or decremented with the defined delta value.
-				The value can be only a positive integer number, otherwise it will be set as 1, or in the cases with double or float the the whole part will be taken.
+				If spinDelta is defined as an intereger, then this value it is applied to all time periods - years, days, minutes, etc.
```
damyanpetev: @Lipata Can you split those into separate descriptions for number and object (enum-like)?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7788,21 +7788,37 @@
 				spin type="string" Spin buttons are located on the right side of the editor
 			*/
 			buttonType: "none",
-			/* type="number" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
+			/* type="number|object" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
 				When not editing (focused) the delta is applied on the day if available in the input mask or the lowest available period.
 				When in edit mode the time period, where the cursor is positioned, is incremented or decremented with the defined delta value.
-				The value can be only a positive integer number, otherwise it will be set as 1, or in the cases with double or float the the whole part will be taken.
+				If spinDelta is defined as an intereger, then this value it is applied to all time periods - years, days, minutes, etc.
+				If spinDelta is defined as an object, then each time period can be defined with specific values. If in that case, some of the time periods don't have values, 1 is set as default one.
```
damyanpetev: If spinDelta is defined as an object, then specific values can be defined for each time period. Time periods that don't have values use 1 as default.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7788,21 +7788,37 @@
...
 			*/
 			buttonType: "none",
-			/* type="number" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
+			/* type="number|object" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
 				When not editing (focused) the delta is applied on the day if available in the input mask or the lowest available period.
 				When in edit mode the time period, where the cursor is positioned, is incremented or decremented with the defined delta value.
-				The value can be only a positive integer number, otherwise it will be set as 1, or in the cases with double or float the the whole part will be taken.
+				If spinDelta is defined as an intereger, then this value it is applied to all time periods - years, days, minutes, etc.
+				If spinDelta is defined as an object, then each time period can be defined with specific values. If in that case, some of the time periods don't have values, 1 is set as default one.
+				The value of each deltas can be only a positive integer number, otherwise exception will be thrown, or in the cases with double or float the the whole part will be taken.
```
damyanpetev: "Accepted values for deltas are positive integer numbers, and the fractional portion of floating point numbers is ignored"
This should stay as a general note I think.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8081,6 +8095,44 @@
...
+					this._validateSpinSettingsPeriod.apply(this, deltaRanges[ index ]);
+				}
+			} else {
+				throw new Error(this._getLocaleValue("spinDeltaIsOfTypeNumberOrObject"));
+			}
+		},
+		_validateSpinSettingsPeriod: function(name, maxRange) {
+			var delta = this.options.spinDelta[ name ];
+
+			if (delta === undefined) {
```
damyanpetev: This can be skipped entirely by not validating undefined values at all (skipping other checks) and simply returning a default 1 in `_getDelta`?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8132,7 +8184,23 @@
 			this._spin(this.options.spinDelta, true);
 		},
 		_handleSpinDownEvent: function () { // DateEditor
-			this._spin(-this.options.spinDelta, true);
+			this._spin(this._setNegativeDelta(this.options.spinDelta), true);
+		},
+		_setNegativeDelta: function (delta) {
```
damyanpetev: 😦 Not super fond of this - the delta shouldn't need to be passed like that down the chain like that, especially with the new `_getDelta`. 
I'm okay with setting it aside till we have the time to redactor though.

Lipata: This means that down the chain we should keep note if we are spinning down or up, which is something I want to avoid. I don't think this will create some performance issues, knowing the fact that the spinDelta object contains 7 properties, so I suggest leave it that way.

ig-ealbert:
@Lipata Please use the following translations for Japanese.
spinDeltaIsOfTypeNumberForPeriod: "{0} に異なる spinDelta タイプが必要です。{1} ～ {2} の正の数を入力してください。",
spinDeltaIsOfTypeNumberOrObject: "異なる spinDelta タイプが必要です。他の時間間隔のデルタを定義する正の数またはオブジェクトを入力してください。",
spinDeltaShouldBeInRange: "{0} の spinDelta オプションは {1} から {2} の間にする必要があります。",

src/js/modules/infragistics.ui.editors.js
```diff
@@ -7788,21 +7788,37 @@
...
+				//Initialize with number
 				$(".selector").%%WidgetName%%({
 					spinDelta: 10
 				});
+				//Initialize with object
+				$(".selector").%%WidgetName%%({
+					spinDelta: {
+						years: 4,
+						months: 3,
+						days: 10,
```
IG-Stanimir-Dimitrov: Agreed - to finalize the debate here is object properties naming convention:

```
spinDelta: { 
     year: 4, 
     month: 3, 
     day: 10, 
     hours: 12, 
     minutes: 15, 
     seconds: 10, 
     milliseconds: 100
}
```

dafo:
1. Run a date  picker: 
```
$("#editor").igDatePicker({
        value: new Date(2017, 11, 8),
        dataMode: "date",
        dateInputFormat: "dateTime",
	spinDelta: {
	        year: 4,
	        month: 3,
		day: 10,
		hours: 12,
		minutes: 15,
		seconds: 10,
		milliseconds: 100
	}
    });
```
2. Click inside the "hours" section
3. Start spinning up
4. Start spinning down

Actual result:
Spinning down is not possible

Expected result:
Spinning up/down works correctly 

Note: If you start spinning down initially, spinning down works fine but spinning up is not working.

dafo:
1. Run a date picker:
```
$("#editor").igDatePicker({
        value: new Date(2017, 11, 8, 01),
        dataMode: "date",
        dateInputFormat: "dateTime",
	spinDelta: {
	        year: 4,
	        month: 3,
		day: 10,
		hours: 12,
		minutes: 15,
		seconds: 10,
		milliseconds: 100
	}
    });
```
2. Click inside the "minutes" section
3. Spin down a single time

Actual result:
The date is changed to 12/**07**/2017 12:45 AM

Expected result:
The date is 12/**08**/2017 12:45 AM

PR #1139, opened 2017-08-18
Allow igloader to load multiple locale and regional scripts

MayaKirova:
Allow igloader to load multiple locale and regional scripts


src/js/infragistics.loader.js
```diff
@@ -2170,15 +2171,23 @@ $.extend($.ig.loaderClass, {
 				if (this._resources[ i ].locale) {
 					localeScripts = this._resources[ i ].locale.slice(0);
 					while (localeScripts.length > 0) {
-						scriptData.unshift(localeScripts.pop().replace("$locale$", this.settings.locale));
+						locales = this.settings.locale.split(",");
+						localeScr = localeScripts.pop();
+						for ( k = 0; k < locales.length; k++) {
+							scriptData.unshift(localeScr.replace("$locale$", $.trim(locales[ k ])));
```
MayaKirova: I don't see a problem. 

This is the current code:
localeScripts = this._resources[ i ].locale.slice(0);
while (localeScripts.length > 0) {
...
localeScr = localeScripts.pop();
...
}

Slice(0) just copies the array ( from beginning to end) so we'll loop through all items in the array as expected.

PR #1140, opened 2017-08-18
Editors locale and regional changes

Lipata:
In addition to the changes in editors the logic that format a date has been moved to utils.jquery. The function is $.ig.formatDate = function (mask, date, regional). This is because no such functionality is presented in the util functions (formatter is not working for editors). We should adapt the grid to use that new format function.


src/js/modules/infragistics.ui.editors.js
```diff
@@ -5327,8 +5317,9 @@
 		},
 		_getDisplayValue: function () { //Numeric Editor
 			var value = this._valueInput.val(),
-				decimalSeparator = this.options.decimalSeparator, decimalPoint = ".",
-				minDecimals = this.options.minDecimals, dataMode = this.options.dataMode,
+				decimalSeparator = this._getOptionOrRegionalValue("decimalSeparator"), decimalPoint = ".",
+				minDecimals = this._getOptionOrRegionalValue("minDecimals"), dataMode = this.options.dataMode,
+				displayFactor = this._getOptionOrRegionalValue("displayFactor"),
```
bazal4o: is display factor part of numeric editor. As far as I remember it's part of Percent Editor. If so - it should be moved to percent editor logic. If no ignore my comment.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -8404,15 +8361,14 @@
 		// This method is used to get indices of the date groups within the mask and to convert the date mask into a mask with digit flags valid for igMaskEditor
 		_convertDateMaskToDigitMask: function (mask) {
 			var x, i, j, flag = -1, txt = "", maskVal = mask;
-			if (!maskVal) {
-				maskVal = "";
-			}
-			maskVal = maskVal.replace("dddd", "ddd").replace("ddd,", "").replace("ddd ", "")
```
bazal4o: Is that block moved somewhere?

src/js/modules/infragistics.ui.editors.js
Lipata: Maybe that's correct, but that refactoring is not related to the current issue. I will leave it that way. The problem in the other comment is fixed.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -1813,6 +1796,9 @@
 			this._superApply(arguments);
 			this._changeLocaleForNotifier();
 		},
+		changeRegional: function() {
```
damyanpetev: @Lipata This should probably be in numeric (since text is not affected by regional changes)?
Also should it be unconditional displayValue? What about edit mode?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -4751,18 +4723,15 @@
...
 				break;
-			case "regional":
-				this.options[ option ] = prevValue;
-				throw new Error(this._getLocaleValue("setOptionError") + option);
 
 			case "excludeKeys":
 			case "includeKeys":
 				this.options[ option ] = prevValue;
-				throw new Error(this._getLocaleValue("numericEditorNoSuchOption"));
+				throw new Error(this._getLocaleValue("setOptionError") + option);
```
damyanpetev: I don't think this should be changing though, as the numerics intentionally throw a different error (since they don't expose those options at all)

src/js/modules/infragistics.ui.editors.js
```diff
@@ -10857,6 +10549,11 @@
 			this._skipMessages = false;
 			return valid;
 		},
+		_deleteInternalProperties: function () {
```
damyanpetev: This one seems unnecessary.

PR #1149, opened 2017-08-24
Task 241535: Implement global change regional and make changeRegional method public

MayaKirova:
Task 241535: Implement global change regional and make changeRegional method public


MayaKirova:
It seems to be setting $.ig.regional.defaults, which is different than changing the regional runtime. Why should we deprecate it?

PR #1152, opened 2017-08-28
Implementing checkbox and expanded bindings

### LINKED ISSUE #1130: In igTree there is no checkbox state binding
### Description  
I want to have the functionality to load the checkbox state from the data source.

### Steps to reproduce  

I need a new binding option "checkedKey".

kdinev:
Closes #1130 

The pull request is to track progress on the item. I'm committing a failing rendering unit test first and then I will proceed with implementation.


kdinev:
@mpavlinov @damyanpetev I'm done with the implementation. Please review and test.

src/js/modules/infragistics.ui.tree.js
```diff
@@ -1978,8 +1996,10 @@
 				opt.bindings.childDataProperty = "Nodes";
 				schema.target = { name: "Target", type: "string" };
 				opt.bindings.targetKey = "Target";
-				schema.expanded = { name: "Expanded", type: "boolean" };
-				opt.bindings.expandedKey = "Expanded";
+				schema.expanded = { name: "__expanded__", type: "boolean" };
+				opt.bindings.expandedKey = "__expanded__";
+				schema.checked = { name: "__checked__", type: "boolean" };
```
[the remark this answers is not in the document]
damyanpetev: And since I can't add a comment on code outside the change - there's a case with `if (bindings.expandedKey) {` at L2036 that adds `schema.fields.push({ name: "Expanded", type: "boolean" });` to the schema rather than the actual key. And there isn't one for `checkedKey` at all.
Also since there's no XPath flavor for expanded and checked options, should those be working with XML binding at all?

src/js/modules/infragistics.ui.tree.js
```diff
@@ -3678,155 +3804,114 @@
...
+				}
+				if (event) {
+					this._triggerNodeCheckstateChanged(event, nodeObject);
 				}
-				self._triggerNodeCheckstateChanged(event, node);
 			}
 		},
-		toggle: function (node, event) {
-			/* Toggles the collapse/expand state for the specified node.
+		checkNode: function (nodeObj, cascadeDir) {
```
damyanpetev: Is this intended to a new public method? If so, the `nodeObj` param is very tricky to create for a mere mortal - 
the format is not really described and the `_constructNodeObject` is kinda private (events expose node objects, but that's not reasonable to go for).
This should probably be folded under a `_checkNode` with the public method accepting a DOM elements as the "node element" description implies based on other methods.
This also applies to `uncheckNode` and `partiallyCheckNode` ofc.

kdinev: The public API that should be used to get the node objects is [`nodeFromElement`](https://www.igniteui.com/help/api/2017.1/ui.igtree#methods:nodeFromElement)

damyanpetev: Right, missed there's a public method that does the exact same thing :)

PR #1182, opened 2017-09-12
Preserve dateInputFormat and dateDisplayFormat options

### LINKED ISSUE #1180: igDatePicker does not change its regional when initialized without options
### Description  
Init an igDatePicker with no options - `$("#dp").igDatePicker();`

### Steps to reproduce  

1. Change its regional option: `$("#dp").igDatePicker("option", "regional", "ja");`

### Result  
It does not change the regional settings

### Expected result  
igDatePicker should apply its new regional settings when initialized without options

Lipata:
Closes #1180

Commit only tests.


Lipata:
The coverage is down, but the modified method [_applyRegionalSettings](https://coveralls.io/builds/13229128/source?filename=src%2Fjs%2Fmodules%2Finfragistics.ui.editors.js#L7981) is fully covered.

damyanpetev:
I've added changes for the actual picker update of DatePicker regional change that was missed in #1140.
Also found out the combined regional doesn't carry the picker options at all -> #1186
Guessing now @bazal4o will have to do the review 😈 
// CC @dkamburov for tracking

PR #1185, opened 2017-09-12
Fix grid styles for all non-bootstrap themes.

desig9stein:


kdinev:
@desig9stein I'm pretty sure the dialog is not border-box. You can try fixing the issue by adding it!

PR #1197, opened 2017-09-15
Removing the encode method from the igTemplating

### LINKED ISSUE #1177: Remove "encode" function in the templating engine
### Description  
With the fix for TFS 238125 and as suggested by @kdinev a new $.ig.encode function is defined in infragistics.util.jquery.js file( see #1125 , #1126 , #1129 ). In the infragistics.templating.js file there is the same function which overrides the one in infragistics.util.jquery.js file.

### Steps to reproduce  
1. Put a breakpoint in $.ig.encode function in both files
2. Execute `$.ig.formatter("<div></div>", "string", "checkbox", true, undefined, undefined, undefined, "<div></div>")`
3. Observe which breakpoint is hit

### Result  
The breakpoint in infragistics.templating.js is hit.

### Expected result  
The breakpoint in infragistics.util.jquery.js should be hit.

### Attachments  
Attach a sample if available, and screenshots, if applicable.

kdinev:
Closes #1177

Adding a dependency on `util.jquery` in `templating`.

@mpavlinov Btw, the encode method doesn't use `jQuery`, and I don't know why it's being put in the `jQuery` file. In fact, quite a few of the new methods I saw there are the same way - no dependency on `jQuery`.

### DIFF
--- a/src/js/modules/infragistics.templating.js
+++ b/src/js/modules/infragistics.templating.js
@@ -9,6 +9,7 @@
  *
  * Depends on:
  *	jquery-1.9.1.js
+ *  infragistics.util.jquery.js
  */
 
  /*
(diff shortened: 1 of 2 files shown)


mpavlinov:
@kdinev Agree. There are a lot of functions that should be moved from infragistics.util.jquery.js to infragistics.util.js. I created a TFS task to address that.

PR #1226, opened 2017-09-26
Fix for bug #242699

ChronosSF:
Changing window.$ or global $ to jQuey for noConflict support.

The bug cannot be automated as part of the loader tests as it needs a clean window state.


ChronosSF:
@kdinev, @pzhekov I'll wait for the AMD fixes for the regional files that @Lipata is doing as those are required for the $.noConflict support as well. The issue is not at all limited to util for the master branch as it was for 17.1.

mpavlinov:
@ChronosSF I guess you're referring to #1240?

PR #1240, opened 2017-10-02
AMD Changes

Lipata:

All widgets should return $, instead of the control itself in order to propagate the jQuery object. All localization files depend on jQuery.


src/js/modules/i18n/infragistics.ui.zoombar-ru.js
```diff
@@ -9,14 +9,13 @@
 
 (function (factory) {
 	if (typeof define === "function" && define.amd) {
-		define( [], factory );
+		define( ["jquery"], factory );
 	} else {
-		return factory();
+		factory(jQuery);
```
kdinev: There should be a `return` because if the factory isn't returning, then doing  

`var locale = require("path/to/locale");`

will result in `undefined`

src/js/modules/infragistics.util.jquerydeferred.js
```diff
@@ -29,7 +29,7 @@
 	} else {
 
 		// Browser globals
-		factory(jQuery);
+		return factory(jQuery);
```
kdinev: Nothing is being returned here.

src/js/modules/infragistics.util.jquery.js
```diff
@@ -28,7 +28,7 @@
 	} else {
 
 		// Browser globals
-		factory(jQuery);
+		return factory(jQuery);
```
kdinev: Nothing is being returned here.

src/js/modules/infragistics.util.js
```diff
@@ -29,7 +29,7 @@
 	} else {
 
 		// Browser globals
-		factory();
+		return factory();
```
kdinev: Here the return is OK, but remove the `// Browser globals` comment, since the module actually returns `igRoot`.

Issue #1179, opened 2017-09-07
knockoutjs issue with requirejs amd module


seawerst:
In your knockout extensions, you define a dependency on knockout but it's not passed to the definition function as function argument. Instead you only enter a comment `/*global ko*/`. How do you manage to have ko globally defined when knockout also `define(...` when it detect requirejs so it's no longer in global.

You should probably use something like:
`(function (factory) {
	if (typeof define === "function" && define.amd) {
		define( [
            "jquery",
            "knockout",
            "jquery-ui",
            "../modules/infragistics.util",
			"../modules/infragistics.util.jquery",
			"../modules/infragistics.util.jquerydeferred",
            "../modules/infragistics.ui.editors"
            ], factory );
	} else {
		factory(jQuery, ko);
	}
}
(function ($, ko) {
	function updatePrope`

![image](https://user-images.githubusercontent.com/5280138/30170825-281f4aba-93be-11e7-84db-1b18c3befcb9.png)

Notice the ordre of the dependencies and the additional function parameters.

Lipata:
@seawerst thank you! Just FYI the  `/*global ko*/` is related to the JSHint validation for our code and it says that `ko` is globally defined.

PR #1243, opened 2017-10-02
#1179 Add knockout as a parameter

Lipata:
#1179

Add knockout as a parameter that is passed to the extension. Remove unnecessary dependencies.

### DIFF
--- a/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
+++ b/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
@@ -16,17 +16,16 @@
 	if (typeof define === "function" && define.amd) {
 		define( [
             "jquery",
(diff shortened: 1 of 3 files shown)


src/js/extensions/infragistics.ui.combo.knockout-extensions.js
```diff
@@ -16,17 +16,16 @@
 	if (typeof define === "function" && define.amd) {
 		define( [
             "jquery",
-            "jquery-ui",
             "knockout",
             "../modules/infragistics.util",
             "../modules/infragistics.datasource",
             "../modules/infragistics.ui.combo"
```
kdinev: I believe that the combo already has dependency on the datasource and the util. Could you check that also for tree and editors?

PR #1258, opened 2017-10-03
Adding ossScriptPath for multi-path loading (TFS Git + submodule scenario)

ChronosSF:
This is not related to a GitHub issue but rather an exploratory fix for the current problem when trying to use the loader in the split TFS Git + submodule directory configuration.

There can be multiple ways to resolve this. A simpler fix without the need to flag the OSS files would produce file not file errors in the browsers console. The idea is also that this doesn't affect the production scenario (all files in one dir) but there is technically one extra condition that needs to be executed for each file so there will be minor performance degradation for the loader.

Using the PR-ed branch you could load files like this avoiding the need to manually copy paste all files in a non-Git dir:

$.ig.loader({
	scriptPath: "http://localhost/ignite-ui/js/",
	ossScriptPath: "http://localhost/ignite-ui-oss/js/",
	cssPath: "http://localhost/ignite-ui-oss/css/"
});

If you alias ignite-ui-oss to point to ignite-ui/modules/igniteui/ or just the GitHub repo.

Mostly looking for feedback. We don't need to merge this right away but it's definitely a problem that's worth looking at from productivity standpoint.

### DIFF
--- a/src/js/infragistics.loader.js
+++ b/src/js/infragistics.loader.js
@@ -145,6 +145,7 @@ $.ig.dependencies = [
 		widget: "theme",
 		scripts: [  ],
 		internal: true,
+		oss: true,
 		css: [ "$path$/themes/$theme$/infragistics.theme.css" ],
 		description: $.ig.loaderClass.locale.descriptions.themeDescription
 	},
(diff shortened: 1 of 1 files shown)


kdinev:
This is related to internal development and testing and should not appear as a public configuration. Instead the loader should be pointed to the `dist/` folder and testing should be performed after `grunt build`.

PR #1259, opened 2017-10-03
 add standard non-token format strings to date formatting

davidnegley:
Closes tfs 243378


davidnegley:
formatToParts is an experimental feature, can't use this

PR #1260, opened 2017-10-03
add standard datetime formats

davidnegley:
Closes tfs 243378


davidnegley:
what's the deal with the "can't find variable" error running the tests?

kdinev:
Hi @davidnegley 
I pulled your branch and ran the tests and they fail on my end as well.

The issue seems to be here:

`if (Intl) {`

Should be `if (window.Intl)`.

src/js/modules/infragistics.util.js
```diff
@@ -1537,6 +1537,13 @@
 
 			return result;
 		};
+		var applyFormat = function(options) {
+			if (Intl) {
```
kdinev: This should be referenced as `if (window.Intl) {`

PR #1282, opened 2017-10-05
Fix header caption to use correct variable for color in all themes

### LINKED ISSUE #1277: Header caption background is white in the Superhero theme
### Description  
The header caption background is white in Superhero theme.

### Steps to reproduce  

1. Open http://jsfiddle.net/n1q9mtpb/3/
2. Observe the header caption

### Result  
caption

### Expected result  
caption_expected

The header caption background should be blue.

desig9stein:
Closes #1277


skrustev:
I noticed that the header caption color in the Infragistics theme is white, while in 17.1 is dark grey like the column header color. I assume this is intentional?

desig9stein:
We have that fixed here https://github.com/IgniteUI/ignite-ui/pull/1283

PR #1283, opened 2017-10-05
Comment out the rules that cause the problem

### LINKED ISSUE #1274: there are classes that look for files that cannot be found
### Description  
There are classes in jquery-ui.css that look for files that cannot be found.
e.g.
.ui-icon,
.ui-widget-content .ui-icon {
    background-image : url("images/ui-icons_444444_256x240.png");
}

.ui-widget-header .ui-icon {
    background-image : url("images/ui-icons_444444_256x240.png");
}

These files cannot be found in their respective paths leading to browsers not applying the rule:

capture
(no rule overrides this, the file is just not found)

And more importantly lead to errors in node apps using package managers - for example:

capture2

There are probably other files that are missing. All these issues need to be resolved in one way or another.

desig9stein:
Closes #1274


src/css/structure/jquery-ui.css
```diff
@@ -1126,7 +1126,8 @@ a.ui-button:focus {
     height : 16px;
 }
 
-.ui-icon,
+
+/*.ui-icon,
```
kdinev: The issue comes from the existence of this file and the fact that it's being combined with the rest of the structure files. I believe it should be removed entirely.

kdinev: I think the change should be made to the `jquery-ui-overrides.less` instead of touching the jQuery UI theme directly. If the jQuery UI css is updated, the commented out parts will come back.

desig9stein: Sure this makes sense.

dkamburov:
FeatureChooser icons are too small and filtering editors don't have icons
image

ChronosSF:
The theming changes allow for the correct compiling of node apps created with create-react-app.

desig9stein:
@dkamburov  I've made another commit which fixes the problem with icons size.

dkamburov:
These are resolved! I'm also approving this.

PR #1302, opened 2017-10-10
Fix TFS bugs for Checkboxes

desig9stein:


src/css/themes/ios/infragistics.theme.css
```diff
@@ -2246,7 +2246,7 @@ body .ui-tooltip {
 .ui-igcheckbox-normal .ui-icon,
 .ui-iggrid th .ui-igcheckbox-normal.ui-state-default .ui-icon,
 .ui-igcombo-checkbox .ui-icon {
-  background-image: url(images/ui-icons_007aff_256x240.png);
+  color: url(images/ui-icons_007aff_256x240.png);
```
mpavlinov: @desig9stein This is invalid CSS.

desig9stein: It should be ok now

PR #1328, opened 2017-10-16
Fix "Click on the filtering editor does not enter in edit mode"

desig9stein:
Closes TFS bug 237522


MayaKirova:
It still doesn't fully span to the width of the container. The following class seems to be applied that does not seems to be needed:
.ui-iggrid-filtercell .ui-igedit-container > .ui-igeditor-input-container {
    float : left;
    width : calc(100% - 10px);
}
It's removed in 17.1.

image

PR #1335, opened 2017-10-17
Fix for jQueryUI context menu 17.1

desig9stein:


src/css/themes/bootstrap3/yeti/infragistics.theme.css
```diff
@@ -1149,7 +1149,8 @@ body .ui-tooltip {
 .ui-icon-carat-1-ne {
   background-position: -16px 0;
 }
-.ui-icon-carat-1-e {
+.ui-icon-carat-1-e,
```
kdinev: Why is this necessary? Is it a workaround for something?

desig9stein: the class “ui-icon-caret-1-e” in the new version of jQueryUI it’s different from the one “ui-icon-carat-1-e” in the old version. 
Now I know it seems odd but there is actually a different ASCI symbol in the class from the new version

kdinev: OMG. I didn't see the `carAt` vs. `carEt`.

Lipata:
Taking the latest SR for 17.1 (2029) and using the ccs folder with the changes from this pull requests adds the arrows for the sub-menu, but there is another visual problem:

image

src/css/themes/bootstrap3/yeti/infragistics.theme.css
desig9stein: lol, you are not the only one :)

PR #1376, opened 2017-11-08
Fix changeGlobalLanguage code sample

dkamburov:


mpavlinov:
This is already fixed.

PR #1388, opened 2017-11-23
FIX-TFS-246065-17.1

desig9stein:
FIX "Grid's header and body is misaligned when a grid is Multi-Row Layout and is continuous virtualization mode."


MayaKirova:
Hi Marin,

I tested the latest changes however they apply double borders on the normal (non-multi-row) grid, while before there we no borders.
comparison

The borders should probably be applied only for multi-row-grid ( should not overwrite the default grid borders with !important).

ChronosSF:
@kdinev, @desig9stein more commits have been pushed to this PR after the bug has been verified. This is not acceptable.

17.1 currently has a breaking visual change in the infragistics theme that adds visible borders to the cells. This needs to be resolved before the upcoming SR.

capture

I've opened issue #1435 for this.

kdinev:
@ChronosSF Unfortunately, I've overlooked the commit that got into after the verification.

PR #1389, opened 2017-11-23
FIX-FOR-246065-17.2

desig9stein:


tbelcheva:
Grid has no visible borders on some of the cells (the borders are visible only on hover). Attached is image for demonstration.
noborder

desig9stein:
@tbelcheva Can you please attach the same sample as the one from your screenshot.

tbelcheva:
Attaching the sample.

[igGrid-reproducible-sample.zip](https://github.com/IgniteUI/ignite-ui/files/1538633/igGrid-reproducible-sample.zip)

PR #1390, opened 2017-11-23
FIX-TFS-246065

desig9stein:
FIX "Grid's header and body is misaligned when a grid is Multi-Row Layout and is continuous virtualization mode."


tbelcheva:
The same behavior as in 17.2: https://github.com/IgniteUI/ignite-ui/pull/1389

PR #1394, opened 2017-11-28
Set text-overflow: ellipsis; to row selector numbers, fixing internal…

dkamburov:
… bug #244284

Closes #244284


desig9stein:
Hi @dkamburov, 
looks OK, but you need to add "white-space: nowrap;" to make sure that the text will never go in two rows.

PR #1402, opened 2017-11-30
Removing the infragistics.util.jquerydeferred.js dependency

DDavidkov:
Remove the infragistics.util.jquerydeferred.js dependency from ignite-ui as it is no longer necessary. Also reflected this in the loader.


kdinev:
@DDavidkov Can you merge IgniteUI/master into your branch?

DDavidkov:
@kdinev OK. This should be done. My fork and IgniteUI/master are synced.

gmurray81:
Did you guys re-route what the Task class talks to which this change? Or was this a straight up polyfill that can be assumed to be present in the min jQuery?

Issue #1403, opened 2017-11-30
igTextEditor dropdown container does not render with correct height


tbelcheva:
[test.zip](https://github.com/IgniteUI/ignite-ui/files/1518721/test.zip)
### Description  
When you have igTextEditor with dropdown mode and you open it, the respective list items are not fully rendered due to incorrect dropdown container height.

### Steps to reproduce  

1. Run the sample
2. Click to open the dropdown

### Result  
The dropdown container has no correct height

### Expected result  
The dropdown container has correct height

PR #1407, opened 2017-12-04
Fixing issue 1403 + tests - 17.1

dafo:
#1403

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -2362,7 +2362,7 @@
 		},
 		_renderList: function () {
 			var i, list = this.options.listItems, itemValue, currentItem, itemHeight, dropdown,
-				id = this.id, html;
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -2398,8 +2398,8 @@
 				itemHeight = 23;
 			}
 			if (list.length < this.options.visibleItemsCount) {
-				dropdown.css("height", parseFloat(itemHeight * list.length));
-				this._listInitialHeight = parseFloat(itemHeight * list.length);
+				dropdown.css("height", parseFloat((itemHeight * list.length) + 2));
```
bazal4o: Here I found code left by me (with magic numbers) - I suggest to improve it https://github.com/bazal4o/ignite-ui/tree/tpaskalev/issue_1403
If you are ok - we can keep the tests and even add more related to the item border (set to more than 1px) and even dropdownlist container border set to more than 1.

PR #1422, opened 2017-12-08
notify "chart" on resize, even if there is no width or height property on "chart"

davidnegley:
Closes tfs 245174


kdinev:
@davidnegley JSCS is failing with spaces instead of tabs.

PR #1428, opened 2017-12-12
Cell value overlaps when editorOption.disabled is enabled via columnSettings

desig9stein:
Fix for TFS Bug 220509


damyanpetev:
@desig9stein As was in the bug note, I believe the [`.ui-iggrid .ui-igedit-disabled`](https://github.com/desig9stein/ignite-ui/blob/279db70214a2c2e92d53e3e9d92c8891dd340fb9/src/css/structure/modules/infragistics.ui.grid.css#L1485) rule does the same thing already. Just needed that one updated as it uses the custom `ui-igedit-disabled` class from editors pre-15.2.
So maybe reuse that rule or delete the obsolete part of it?

src/css/structure/modules/infragistics.ui.grid.css
```diff
@@ -2124,4 +2124,13 @@ margin:1em 20%;
 
 .ui-iggrid-featurechooser-list-submenu li.ui-state-hover {
     background:#eee;
-}
\ No newline at end of file
+}
+
+/* BUG 220509 11.12.2017 Marin Popov */
+.ui-widget .ui-igedit.ui-state-disabled {
```
damyanpetev: This applies to both rules - scope them to `.ui-iggrid` or something similar, otherwise they apply outside the grid which is not intended.

mpavlinov:
@desig9stein The issue is still reproducible. Checkout the screenshot:
image

desig9stein:
@mpavlinov sorry about that, I committed the wrong selector, now it should be fine.

mpavlinov:
@desig9stein Can you make PRs for 17.2 and 18.1 please.

PR #1439, opened 2017-12-13
in $.ig.util.stringToColor, return transparent for null string

davidnegley:
tfs 247152 (also fixed in calling code, just adding this failsafe for ig_util)


kdinev:
@davidnegley Could you include a few unit tests for this?

PR #1443, opened 2017-12-14
[igCombo] Items are not correctly rendering when the data source items are less than the Combo visible items count with grouping and virtualization enabled

### LINKED ISSUE #1441: [igCombo] Items are not correctly rendering when the data source items are less than the Combo visible items count with grouping and virtualization enabled
### Description  
List items are not rendered correctly when the items in the data source of the combo are less than the combo `visibleItemsCount` property with grouping and virtualization enabled. As a bonus filtering through the combo input will start removing and item from the dropdown for each filter operation until the dropdown is empty.

### Steps to reproduce  
1. Download and open the sample
2. Open the dropdown. The amount of rendered items does not match the data source items.
3. Start filtering and clearing the input of the combo (3-5 times)

### Result  
The amount of items rendered gets progressively fewer until the combo dropdown no longer renders.

### Expected result  
The combo renders the correct number of items and should not trim items after each filtering operation.

### Attachments  
[combogroup.zip](https://github.com/IgniteUI/ignite-ui/files/1559330/combogroup.zip)

rkaraivanov:
Closes #1441


src/js/modules/infragistics.ui.combo.js
```diff
@@ -2163,7 +2162,10 @@
         _groupsMarkup: function () {
             var groups, groupsLen, i,
 				dataView,
-				dataLen = this._itemsToRenderCount(),
+                dataLen = this._itemsToRenderCount(),
```
kdinev: Here spaces should also be tabs.

Issue #1435, opened 2017-12-13
igGrid has borders after latest CSS changes in 17.1


ChronosSF:
### Description  
The grid has visible borders which is a breaking visual change for 17.1

### Steps to reproduce  

1. Open the attached sample.
2. Observe the borders

### Result  
The grid has visible borders between columns

### Expected result  
The grid should not have visible borders so that the visual design is consistent with 17.1 RTM

### Attachments  
![capture](https://user-images.githubusercontent.com/2188411/33940210-ccbe3064-e016-11e7-93d0-b9ec5df88eec.JPG)
[amd_test.zip](https://github.com/IgniteUI/ignite-ui/files/1555561/amd_test.zip)

PR #1449, opened 2017-12-18
Fix grid double borders 

desig9stein:


MayaKirova:
Verified that tfs tests for grid and treegrid pass when run against these changes.

ChronosSF:
Closes #1435

ChronosSF:
@desig9stein the only issues i can see with this PR are:

- Borders appearing when hovering in bootstrp4
bootstrap4
- Borders with ios (there are no borders with ios theme in 16.2 so I guess we have to fix this too)
ios

PR #1451, opened 2017-12-18
Revert transparent borders

desig9stein:
Closes #ISSUE_NUMBER


ChronosSF:
@desig9stein this differs from 17.1's behavior and I think it regresses a previous fix in the multi-row-layout scenario.

In 17.1 the following style is used in multi-row-layout grid:
.ui-iggrid .ui-iggrid-table-mrl tbody td,
.ui-iggrid .ui-iggrid-table-mrl tbody td.ui-state-hover {
  border: 1px solid #dadada;
}

It overrides the following because it is more specific (I think)
.ui-iggrid tbody td {
  border-left: 1px solid transparent;
  border-right: 1px solid transparent;
}

And the grid looks nice:
infragistics

In this 17.2 PR the transparent rule overrides whatever is used for MRL and the grid then doesn't have borders for this scenario:
capture

Also there seems to be similar issues in hover styles in Bootstrap4, Yeti and Flatly:
flatly
yeti
bootstrap4

ChronosSF:
@desig9stein  These are the samples I am testing with. Please check my findings.
[samples.zip](https://github.com/IgniteUI/ignite-ui/files/1568366/samples.zip)

PR #1457, opened 2017-12-18
Fix missing styles for active list item

desig9stein:


dafo:
Marin, currently the "active" and "selected" items share the same style (the same shade of blue). The selected item should be a bit darker. 
Note: When the selected item is also the active one - it should receive the "active" style.

mpavlinov:
@desig9stein There are some conflicts. Can you resolve them please.

PR #1463, opened 2017-12-19
Dpetev/datepicker enter reset 1453

### LINKED ISSUE #1453: [igDatePicker]Entered date is converted to today's date when pressing the Enter key
### Description  
Entered date is converted to today's date when pressing the Enter key

### Steps to reproduce  

1. Open attached sample (editor.htm)
2. Open igDatePicker's dropdown
5. Click in the igDatePicker's input area
6. Enter a date different from today's date
7. Press the "Enter" key

### Result  
The date is converted to today's date

### Expected result  
The entered date should be kept.

### Attachments  
[editor.zip](https://github.com/IgniteUI/ignite-ui/files/1568115/editor.zip)

damyanpetev:
Closes #1453 

**Many for the price of one**
Changes include:
- DatePicker internal mask using first digit as optional (should't affect any functionality since required validation was not used in the date editors so far)
- Tied `datepicker` sync to `textChanged`-called function and restricted to just when the picker is open, so only adding functionality in this specific case
- Updated `datepicker` instance's property `lastVal` in a few places where we override the input text with our formatting. This makes the jQuery widget change checks pass and skip re-parsing and failing on our format, thus avoiding resetting to the default (today).

The result should be that the input of our editor should properly update the jQuery date picker when filled enough to form a value and work as it would with US format dates.


tests/unit/common/test-util.js
```diff
@@ -105,16 +105,54 @@
 				$input[0].selectionEnd = startPos + newVal.length;
 			}
 		},
+
+		/**
+		 * Returns a new Promise that resoles after a set period of time
+		 * @param {number} ms Amount of ms to wait
+		 */
+		wait: function (ms) {
+			if (false) {
```
bazal4o: I guess we need a check for Promise existence here :)

PR #1467, opened 2017-12-20
#1466 Set properly null value to numeric editor

### LINKED ISSUE #1466: Null value is not properly set when value is deleted from numeric editor
### Description  
When igNumericEditor allow to have null as a value, when you delete existing value in the editor and leave an empty string then value is not properly set to null, but stay as empty string. This is item #246940 in TFS.

### Steps to reproduce 

```
$("#editorNull").igNumericEditor({
    allowNullValue: true,
    nullValue: null,
    value: 0
});
```

1. Run the code above
2. Focus the editor and delete the 0 value and leave ""
3. Blur the editor
4. Execute $("#editorNull").igNumericEditor("value") method 

### Result  
The value method returns "" (empty string)

### Expected result  
The value method returns null.

Lipata:
Closes #1466

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4836,7 +4836,7 @@
 			if (value !== this.value()) {
 
 				// N.A. December 20th, 2017 #665: When new value is empty string we call clearValue(), to properly set new value according to the editor options.
-				if (value === "" ) {
+				if (value === "" && this.options.allowNullValue && this.options.nullValue === null) {
 					this._clearValue();
 				} else {
 					this._updateValue(value);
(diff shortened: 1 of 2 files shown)


dafo:
@Lipata,
$("#editorNull").igNumericEditor({ allowNullValue: true, nullValue: , value: 0 });

Run the code above
Focus the editor and delete the 0 value and leave ""
Blur the editor
Execute $("#editorNull").igNumericEditor("value") method
Actual: The value method returns "" (empty string)
Expected: The value method should return 5.

Lipata:
@dafo As we discussed it, this will stay in this way, cause empty string is allowed value and igNumericEditor will accept the empty string("") as a NEW value instead of a NULL value - it's the same in the igTextEditor.

PR #1473, opened 2017-12-22
Add minimum size for scrollbars thumb drag. Improve logic regarding thumb drag positioning.

### LINKED ISSUE #1135: IgScroll custom scrollbar thumb gets too small when the scroll container has big height
[defaultTable-igScroll.zip](https://github.com/IgniteUI/ignite-ui/files/1231459/defaultTable-igScroll.zip)

### Steps to reproduce  

1. Open the attached sample
2. Hover over the vertical scrollbar

### Result  
The vertical scroll thumb is very small

### Expected result  
There should be min height of the vertical scroll thumb. Same with the horizontal one.

skrustev:
Closes #1135


mpavlinov:
In Edge and Firefox when I try to drag the thumb with the mouse there is a problem. See the attached video.
[Recording #9.zip](https://github.com/IgniteUI/ignite-ui/files/1661040/Recording.9.zip)

mpavlinov:
@skrustev 
1. Scroll by clicking on an empty area of the scroll
2. Scroll by dragging the thumb

Result
The thumb jumps (see the attached video)
[Recording #10.zip](https://github.com/IgniteUI/ignite-ui/files/1667383/Recording.10.zip)

src/js/modules/infragistics.ui.scroll.js
```diff
@@ -754,6 +754,8 @@
 			this._linkedVBar = null;
 			this._elemWidth = elem.width();
 			this._elemHeight = elem.height();
+			this._minVDragHeightPx = 15; // in px
```
kdinev: Shouldn't this be a setting with a default value of `15`?

skrustev: We decided to not add it as a new option yet but it can easily be exposed.

kdinev: @skrustev It's better to expose it, otherwise this is a "magic number".

skrustev: @kdinev I've exposed it as an option.

Issue #512, opened 2016-11-10
Editors KO extension coverage


kdinev:
Increase the editors knockout extension code coverage to 90%+.

kdinev:
Editors are now at 90.72% so closing the issue.

kdinev:
Ops, this was for editors knockout extension. Reopening.

DDavidkov:
Closing the issue since the coverage is 96.14% now.

PR #1474, opened 2018-01-02
#512 igTextEditor KO extension automation improvement.

gedinakova:
#512 igTextEditor KO extension automation coverage improvements.


tests/unit/knockout/editors/igTextEditor/tests.html
```diff
@@ -337,7 +337,24 @@
 				}, 200);
 			}, 200);
 		});
-
+		var testId = 'Test valueChanged event with non-observable value';
```
kdinev: Don't repeat `var` definitions. Define once and then just assign.

PR #1476, opened 2018-01-02
Remove unnecessary code for saving dialog state

### LINKED ISSUE #1475: Unnecessary code for saving dialog state
### Description  
There is a _save function which only saves some info, but the info is never retrieved.

sstoyanovIG:
Closes #1475

### DIFF
--- a/src/js/modules/infragistics.ui.dialog.js
+++ b/src/js/modules/infragistics.ui.dialog.js
@@ -1149,7 +1149,6 @@
 				elem.hide();
 			}
 			self._created = true;
-			self._save();
 		},
 
 		// get around combination of jQuery with IE7 and other versions of IE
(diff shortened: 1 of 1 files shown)


damyanpetev:
Looking at the hidden input, why do I have the feeling this might be intended for the MVC wrappers for some sort of ill-conceived "state management".. 😞 
Gimme a bit to check it out.

damyanpetev:
@sstoyanovIG Yup, the `inputName` option is only set by the MVC wrappers when you create a Dialog using one of [the overloads with `persistenceInputName` parameter](https://www.igniteui.com/help/infragistics.web.mvc~infragistics.web.mvc.infragisticssuite%601~dialog%28string,string%29).
Guess we are not deleting that bit after all, which makes the PR and issue invalid.
You can still unit-test that is save state separately.

PR #1492, opened 2018-01-04
Zoombar tests update for QUnit2/Karma

ChronosSF:
Updated tests for QUnit2 / Karma runs

Should run fine in the old runner, however, there is a test with 3 script errors in Karma which almost looks like an issue with the webpack bundle somewhere in dv_core as it doesn't reproduce when running the html file or through PhantomJS. This will need further investigation.


damyanpetev:
@ChronosSF I'll take a look, though karma isn't doing anything specific for file loading. Might be double-loaded files and/or order, that setup is still a bit iffy :)

PR #1498, opened 2018-01-05
Fix checkboxEditor tests

wnvko:
Fix checkboxEditor tests


wnvko:
Closed in favor of #1511

PR #1499, opened 2018-01-05
Fixing bug 247461

kacheshmarova:
Fixing bug 247461


tbelcheva:
The error still reproduces. I assume that it's because of the fact that the editor of the first column is undefined and it should be skipped when setting the new language in changeLocale function of igGridFiltering.

PR #1505, opened 2018-01-08
#1417 Fix IME date input for Firefox and IE11

### LINKED ISSUE #1417: Input value is discarded upon lost focus while in composition mode on IE11 and Firefox
### Description  
Input value is discarded upon lost focus while in composition mode on IE11 and Firefox

### Steps to reproduce  
1. Set focus on igDatePicker
2. Enable Japanese IME mode
3. Enter “２０１３０４０４” in composition mode
4. Focus out from igDatePicker

### Result  
Input value is discarded on lost focus.

### Expected result  
Input value is processed on lost focus.

### Attachments  
[index.zip](https://github.com/IgniteUI/ignite-ui/files/1533708/index.zip)

Lipata:
Closes #1417  

In some browsers (like IE 11 and Firefox) blur event is fired before compositionend one, that's why we need to convert the IME input into date inside the _processInternalValueChanging, not in compositionend event handler.
Some refactoring of the composition end event handler.


src/js/modules/infragistics.ui.editors.js
```diff
@@ -2777,7 +2777,7 @@
 				},
 				"compositionend.editor": function () {
 					setTimeout(function () {
-						var value, pastedValue, widgetName = self.widgetName,
+						var pastedValue, widgetName = self.widgetName,
```
bazal4o: it's better to rename the variable from pastedValue - to compositionValue or similar.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -9198,6 +9199,14 @@
 				}
 			}
 		},
+		_parseValueByIMEMask: function() {
```
bazal4o: make the function with parameter - so at least this could be unit tested. Because the

src/js/modules/infragistics.ui.editors.js
```diff
@@ -9178,10 +9174,15 @@
 				}
 				return;
 			}
-			if ($.type(value) === "date") {
-				parsedVal = value;
-			} else {
-				parsedVal = this._parseDateFromMaskedValue(value);
+			parsedVal = value;
```
bazal4o: there is no need of parsedVal at this point. We need to move the definition a few lines down in the if block

src/js/modules/infragistics.ui.editors.js
damyanpetev: Also because hard-coded `this._currentCompositionValue` is not very reusable.

damyanpetev: Actually, I think this should even be part of `_parseDateFromMaskedValue`, just like IME number conversion is part of `_parseNumericValueByMode` :)

damyanpetev: Or part of the respective editors' `_insert`

src/js/modules/infragistics.ui.editors.js
```diff
@@ -6538,19 +6519,25 @@
 			this._promptCharsIndices = [];
 
 			if (this._editMode) {
-				newValue = this._parseValueByMask(newValue);
-				this._editorInput.val(newValue);
-				if (selection !== undefined) {
-					// Move the caret
-					this._setSelectionRange(this._editorInput[ 0 ], selection.start, selection.end);
-				}
+				this._insertEditMode(newValue, selection);
```
bazal4o: If we change the call to _insertEditMode and _insertDisplayMode - we need to be sure we have implementation into the base widgets. If inside a text editor I enter edit mode and type in composition when I submit the input the composition.end handler will try to call those functions - and it will throw an exception.

bazal4o: Ignore my comment

PlamenaMiteva:
While verifying this fix, I found that the dateEditor does not revert the input value when some text (invalid input) is entered and I created a new issue #1525.

PR #1507, opened 2018-01-08
#1417 Fix IME date input for Firefox and IE11

### LINKED ISSUE #1417: Input value is discarded upon lost focus while in composition mode on IE11 and Firefox

Lipata:
Closes #1417  

In some browsers (like IE 11 and Firefox) blur event is fired before compositionend one, that's why we need to convert the IME input into date inside the _processInternalValueChanging, not in compositionend event handler.
Some refactoring of the composition end event handler.

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -2776,7 +2776,7 @@
 				},
 				"compositionend.editor": function () {
 					setTimeout(function () {
-						var value, pastedValue, widgetName = self.widgetName,
+						var pastedValue, widgetName = self.widgetName,
 							cursorPosition = self._getCursorPosition(),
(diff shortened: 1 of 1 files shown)


PlamenaMiteva:
While verifying this fix, I found that the dateEditor does not revert the input value when some text (invalid input) is entered and I created a new issue #1525.

Issue #1083, opened 2017-07-17
Add a new igTimePicker control


IG-Stanimir-Dimitrov:
Create a new component igTimePicker by extending the date editor and adding a dropdown list widget with 24h period split into 30min time intervals.

NishantThakkar:
Any update on the same?

SlavUI:
This will be introduced in Sprint 4

bkulov:
Closing this issue as it's already done. Only MVC wrappers are pending #1563.

PR #1538, opened 2018-01-18
New control - igTimePicker

SAndreeva:
Introduce new control - igTimePicker. #1083 

Implementation and unit tests.


src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,687 @@
...
+				$(".selector").%%WidgetName%%("option", "timeDelta", { minutes: 30 });
+			```
+			object type="object" A configuration object, which defines specific values for each time period. The option can accept the following format:
+				timeDelta: {
+					hours: 0,
+					minutes: 30,
+				}
+			Time periods that don't have values use 0 as default for hours and 30 for minutes.
+			*/
+			itemsDelta: { hours: 0, minutes: 30 },
```
bazal4o: the option is called itemsDelta, while the code snippets are showing timeDelta. either update code snippet, or add the corresponding property.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,687 @@
...
+				});
+
+				//Get
+				var suppressNotifications = $(".selector").%%WidgetName%%("option", "suppressNotifications");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "suppressNotifications", true);
+			```
+			*/
+			suppressNotifications: false,
```
bazal4o: Why do you have those options defined here, when they come from the base widget? Or there is a specific reason which I'm missing?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,687 @@
...
+				});
+
+				//Get
+				var readOnly = $(".selector").%%WidgetName%%("option", "suppressKeyboard");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "suppressKeyboard", true);
+				```
+			*/
+			suppressKeyboard: false,
```
bazal4o: The same as above comment

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+				});
+				```
+				eventArgument="evt" argType="event" jQuery event object.
+				eventArgument="ui.owner" argType="object" Gets a reference to the editor.
+				eventArgument="ui.text" argType="string" Gets a reference to the new text.
+				eventArgument="ui.oldText" argType="string" Gets a reference to the old text.
+			*/
+		},
+		_create: function () { // igTimePicker
+			$.ui.igDateEditor.prototype._create.call(this);
```
kdinev: Use `_super` instead.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+				this._populateDropDown(this.options.minTimeValue, this.options.maxTimeValue);
+			}
+		},
+		getSelectedListItem: function () {
+			/* Gets the selected list item.
+			```
+			var selectedItem =  $(".selector").%%WidgetName%%("getSelectedListItem");
+			```
+				returnType="$" Selected list item.*/
+			return $.ui.igTextEditor.prototype.getSelectedListItem.call(this);
```
kdinev: Use `_super` instead.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+		},
+		dropDownVisible: function () {
+			/* Returns the visibility state of the calendar.
+			```
+			$(".selector").igTimePicker("dropDownVisible");
+			```
+				returnType="bool" The visibility state of the calendar. */
+			return $.ui.igTextEditor.prototype.dropDownVisible.call(this);
```
kdinev: Use `_super` instead.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+		},
+		dropDownButton: function () {
+			/* Returns a reference to the drop-down button UI element of the editor.
+			```
+			var button = $(".selector").%%WidgetName%%("dropDownButton");
+			```
+				returnType="$" Returns reference to jquery object. */
+			return $.ui.igTextEditor.prototype.dropDownButton.call(this);
```
kdinev: Use `_super` instead. Btw, do you really need the overrides, since they do nothing different?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+		},
+		dropDownContainer: function () {
+			/* Gets reference to jquery object which is used as container of drop-down list.
+			```
+				 $(".selector").%%WidgetName%%("dropDownContainer");
+			```
+				returnType="$" Returns reference to jquery object. */
+			return $.ui.igTextEditor.prototype.dropDownContainer.call(this);
```
kdinev: Use `_super` instead. Btw, do you really need the overrides, since they do nothing different?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+		findListItemIndex: function (text, matchType) {
+			/* Finds index of list item by text that matches with the search parameters.
+			```
+				var item =  $(".selector").igTimePicker("findListItemIndex");
+
+			```
+				paramType="string" optional="false" The text to search for in the drop down list.
+				paramType="startsWith|endsWith|contains|exact" optional="true" The rule that is applied for searching the text.
+				returnType="number" Returns index of the found item. */
+			return $.ui.igTextEditor.prototype.findListItemIndex.call(this, text, matchType);
```
kdinev: Use `_super` instead. Btw, do you really need the overrides, since they do nothing different?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,682 @@
...
+		},
+		selectedListIndex: function (index) {
+			/* Gets the index of the selected list item. Sets selected item by index.
+			```
+				$(".selector").%%WidgetName%%("selectedListIndex", 1);
+				var selectedIndex = $(".selector").%%WidgetName%%("selectedListIndex");
+			```
+				paramType="number" optional="true" The index of the item that needs to be selected.
+				returnType="number" Returns the selected index. */
+			return $.ui.igTextEditor.prototype.selectedListIndex.call(this, index);
```
kdinev: Use `_super` instead. Btw, do you really need the overrides, since they do nothing different?

kdinev:
Ignore my previous comments. I actually just noticed what widget the inheritance comes from and why the methods are bypassing the parent widget implementations.

src/js/modules/infragistics.ui.editors.js
sstoyanovIG: I have changed the code snippet

src/js/modules/infragistics.ui.editors.js
sstoyanovIG: This is ignored in the parent control, so it should be added

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+			```
+			object type="object" A configuration object, which defines specific values for each time period. The option can accept the following format:
+				itemsDelta: {
+					hours: 0,
+					minutes: 30,
+				}
+			Time periods that don't have values use 0 as default for hours and 30 for minutes.
+			*/
+			itemsDelta: { hours: 0, minutes: 30 },
+			/* type="number|object" Gets/Sets delta-value which is used to increment or decrement the editor date on spin actions.
```
damyanpetev: ...decrement the editor ~date~ **time** on spin actions

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+				When not editing (focused) the delta is applied on the day if available in the input mask or the lowest available period.
```
damyanpetev: No chance of delta being applied to day, right?

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+			```
+			number type="number" This value it is applied to hours and minutes,.
+			object type="object" A configuration object, which defines specific values for each time period. The option can accept the following format:
+				spinDelta: {
+					hours: 12,
+					minutes: 15
+				}
+			Default value is {hours: 1, minutes: 30}.
+			*/
+			spinDelta: { hours: 1, minutes: 30 },
```
damyanpetev: I know the override is to hide the date parts, but the inherited [`spinDelta`](https://www.igniteui.com/help/api/2017.2/ui.igdateeditor#options:spinDelta) also has seconds and milliseconds. If those still valid in the time picker mask, t maybe the delta options should stay too.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+				});
+
+				//Get
+				var maxTimeValue= $(".selector").%%WidgetName%%("option", "maxTimeValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "maxTimeValue", "18:00");
+			```
+			*/
+			maxTimeValue: null,
```
damyanpetev: Any chance we can merge this with `maxValue` just for consistency?
Also I see the option removed further down, so if this controls just the list items, how do you control the max value if you disable `isLimitedToListValues`?
Seems to make sense for the max value to also limit the list items anyway.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+				});
+
+				//Get
+				var minTimeValue= $(".selector").%%WidgetName%%("option", "minTimeValue");
+
+				//Set
+				$(".selector").%%WidgetName%%("option", "minTimeValue", "09:00");
+			```
+			*/
+			minTimeValue: null,
```
damyanpetev: Same as max

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,695 @@
...
+			/* @Ignored@ Removed from timepicker options*/
+			dateDisplayFormat: null,
+			/* @Ignored@ Removed from timepicker options*/
+			dateInputFormat: null,
+			/* @Ignored@ Removed from timepicker options*/
+			yearShift: null,
+			/* @Ignored@ Removed from timepicker options*/
+			displayTimeOffset: null,
+			/* @Ignored@ Removed from timepicker options*/
+			enableUTCDates: null
```
damyanpetev: This might be okay to leave in, actually. All this option does is to change the default value stored in the submit input, where:
"2018-01-26T13:32:42.540+02:00" would be submitted as "2018-01-26T11:32:42.540Z" instead.
It's really a matter of preference for the dev, depending on the backend, though in most cases it makes no difference.

src/js/modules/infragistics.ui.editors.js
sstoyanovIG: yes

src/js/modules/infragistics.ui.editors.js
sstoyanovIG: I've talked with Slav and we agreed to have only hours and minutes

damyanpetev: Yeah, ignore that one - noticed the s/ms are not part of the input format.

src/js/modules/infragistics.ui.editors.js
sstoyanovIG: Talked with Bobby and we agreed to change this as you are suggesting

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,666 @@
...
+				// In case no propery matches, we call the super. Into the base widget default statement breaks
+				this.options[ option ] = prevValue;
+				this._super(option, value);
+				break;
+			}
+		},
+		_updateValue: function (value) { //igTimePicker
+			if (value === null || $.type(value) === "date" || value === "") {
+				this._super(value);
+				if ($.type(value) === "date" && this._dropDownList !== undefined) {
```
damyanpetev: So if the value is null/empty (thus `$.type(value) === "date"` is false) the dropdown selection won't update?
I think it should always update if the value changes, so this entire block can be simplified with `_super` and `_updateDropdownSelection` after the value handling

damyanpetev: On a second thought, there shouldn't be additional `new Date(value)` parse, because depending on data mode your value could very well be a simple string that's not to be parsed at all.

src/js/modules/infragistics.ui.editors.js
```diff
@@ -12059,5 +12062,666 @@
...
+				}
+			} else {
+				this._super(new Date(value));
+				if (this._dropDownList !== undefined) {
+					this._updateDropdownSelection(new Date(value));
+				}
+			}
+		},
+		_updateDropdownSelection: function (currentVal) { //igTimePicker
+			if ($.type(currentVal) === "date") {
```
damyanpetev: This one shouldn't be required as well - the only calls to this method are from the `_updateValue` above where you control the value and text change that goes through `_valueFromText`, which you should work as is or will need to be modified.

PR #1545, opened 2018-01-22
Move igUpload tests to karma

DDavidkov:
Moving the igUpload tests to karma


damyanpetev:
@DDavidkov Surely you mean `igLoader` tests :)
While the tests are fine, re-loading some of the source files is quite likely to mess up with the rest of the tests. So I'm still holding on merging this, kind of want to come up with a way to skip the actual source loading part - currently I'm thinking of either running these separately or mocking the script load somehow.

tests/unit/loader/loader-test.js
```diff
@@ -0,0 +1,208 @@
...
+	beforeEach: function(){
+		this.cleanScripts();
+	},
+
+	resourcesLoaded: function(expectedResources, assert) {
+		//check if expected resources are loaded
+		var scripts = $(document).find("head > script"), actualPaths = [];
+		scripts.each(function () {
+			//console.log("script.src: " + this.src)
+			if (this.src.indexOf("ignite-ui/src/js/") || this.src.indexOf("igniteui/src/js/") > 0) {
```
damyanpetev: `this.src.indexOf("ignite-ui/src/js/")` doesn't really check anything (kind of why those actually work), so maybe it should use the `jsPath`? Same for the css.

tests/unit/loader/loader-test.js
```diff
@@ -0,0 +1,208 @@
...
+			}
+		});
+
+		for (var i = 0; i < expectedResources.length; i++) {
+			assert.ok(actualPaths.contains(expectedResources[i]), "The loaded resources should contain: " + expectedResources[i]);
+		}
+	},
+	cleanScripts: function() {
+		$(document).find("head > script").each(function () {
+			if (this.src.indexOf("ignite-ui/src/js/") > 0 && this.src.indexOf("src/js/infragistics.loader.js") === -1) {
```
damyanpetev: Similar to the other check, except this one does check the index and doesn't find anything to remove, leaving the scripts/links in place after the test(s)

PR #1546, opened 2018-01-22
Moving templating and tree to karma

kdinev:
Closes #ISSUE_NUMBER


package.json
```diff
@@ -17,13 +17,14 @@
   "license": "Apache-2.0",
   "scripts": {
     "test": "karma start",
-    "cover": "grunt hint && grunt jscs && npm test -- --single-run",
-    "postinstall": "npm install -g grunt-cli"
```
damyanpetev: I'm all for dropping the global postinstall, but we still need grunt.
Could add it Travis scripts and make it required to setup dev env?

kdinev: I'll add it back. I forgot to undo.

PR #1551, opened 2018-01-24
Mvenkov/editors bugs tests

wnvko:
Move editor/bugs test to new Karma/QUnit .
Allowing Enter key to wright '\n' in text area in test-util.keyInteraction()


tests/unit/editors/Bugs/bugs-test.js
```diff
@@ -0,0 +1,1620 @@
+$(document).ready(function () {
```
damyanpetev: The ready function is not really required for these tests anymore so you can remove it :)

PR #1552, opened 2018-01-24
Moving the igSplitter tests to Karma

DDavidkov:
Moving the igSplitter tests to Karma

### DIFF
--- a/tests/unit/common/test-util.js
+++ b/tests/unit/common/test-util.js
@@ -8,6 +8,8 @@
 		checkClass: function (element, cls) {
 			if (QUnit && typeof (QUnit.ok) === 'function') {
 				QUnit.ok(element.hasClass(cls), 'The control with id: ' + element[0].id + ' does not contain the class: ' + cls);
+			} else if (QUnit && QUnit.assert && typeof(QUnit.assert.ok) === 'function') {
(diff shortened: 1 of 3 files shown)


tests/unit/common/test-util.js
```diff
@@ -8,6 +8,8 @@
 		checkClass: function (element, cls) {
 			if (QUnit && typeof (QUnit.ok) === 'function') {
 				QUnit.ok(element.hasClass(cls), 'The control with id: ' + element[0].id + ' does not contain the class: ' + cls);
+			} else if (QUnit && QUnit.assert && typeof(QUnit.assert.ok) === 'function') {
+				QUnit.assert.ok(element.hasClass(cls), 'The control with id: ' + element[0].id + ' does not contain the class: ' + cls);
```
damyanpetev: I think this will just have to get `assert` as parameter, rather than trying the global functions.
Would this even properly count towards the per-test count?

damyanpetev: Indeed, checked and the global `assert` has no test context but it falls back to a stored current so the result gets pushed to the correct test. Could've sworn it failed before, oh well 🤷‍♂️ 
Will take this as it is, though the old version can probably be removed later on.
// CC @wnvko for similar changes

PR #1561, opened 2018-01-26
Fix unable to scroll down using small movement on touchpad in igScroll

### LINKED ISSUE #1438: igScroll doesn't scroll properly in Edge if you scroll slowly with a smooth scrolling trackpad.
### Description  
There is no movement with igScroll if you scroll slowly with a trackpad wheel action, for a smooth scrolling trackpad e.g. the Macbook Pro trackpad.

### Steps to reproduce  

1. Open an igScroll sample in Edge on a machine with a smooth scrolling trackpad.
2. Perform the wheel action slowly (e.g. a 2 finger pan).

### Result  
No scrolling occurs. If you look at the args in a scrolling event, there is a step, but nothing ever changes scrollTop.

### Expected result  
Scrolling should occur.

skrustev:
Closes #1438


src/js/modules/infragistics.ui.scroll.js
```diff
@@ -2161,42 +2163,72 @@
...
+				this._startY = this._getScrollbarVPosition();
+			} else {
+				this._startY = this._getContentPositionY();
+			}
+
+			/** Get delta for the X axis*/
+			if (evt.wheelDeltaX) {
+				/* Option supported on Chrome, Safari, Opera.
+				/* 120 is default for mousewheel on these browsers. Other values are for trackpads */
+				scrollDeltaX = -evt.wheelDeltaX / 120;
```
mpavlinov: @skrustev Move 120 into a variable or constant

src/js/modules/infragistics.ui.scroll.js
```diff
@@ -2161,42 +2163,72 @@
...
 				/* For other browsers that don't provide wheelDelta, use the deltaY to determine direction and pass default values. */
-				scrollDeltaY = evt.deltaY > 0 ? 1 : -1;
+				scrollDeltaY = this._clampAxisCoords(evt.deltaY, -1, 1);
 			}
 
-			if (this.options.smoothing) {
-				//Scroll with small inertia
-				this._smoothWheelScrollY(scrollDeltaY);
+			if (scrollDeltaX) {
+				scrolledX = this._scrollToX(this._startX + scrollDeltaX * scrollStep, true);
```
kdinev: do you need the `scrolledX` assignment as the codacy comment suggests, since it's not used afterwards?

src/js/modules/infragistics.ui.scroll.js
skrustev: There is a comment for that, there's no need for extra variable since extra variable does not describe fully what it is.

gmurray81:
Is clamping it to a min value really a good move, if that's what you are doing? Won't that cause it to desync a bit from the action?

Is there still no zero wheel Delta?

skrustev:
Yes it's pretty much correct that I'm clamping it to a min value. That is pretty much the issue because scrollTop for example takes only round numbers and if you put a float Edge floors that value.

Lets have scrollTop 10px. In the case with the trackpad since the delta is really small it results in scroll ammount of 0.5px for example and when you scroll down you get 10.5px which the browser floors to 10px and doesn't do any scrolling. If you scroll up you get 9.5 which floored is 9px for which the browser reacts and scrolls 1px up. That is why it's needed to be clamped to min value of 1px per scroll event with deltaY.

Why would it cause desync thought? Event if the delta is 0 that's why there is an if statement requiring it to be anything other than 0 in order to clamp it.

PR #1577, opened 2018-02-02
fix data depth 17.2

### LINKED ISSUE #1564: igTree does not load nodes after lv3 when using loadOnDemand
### Description  
igTree does not load nodes after lv3 when using loadOnDemand,  because
_depth = ul.attr("data-depth")_ is not parsed. This leads to setting wrong _attr("data-depth")_ for the nested node levels.

### Steps to reproduce  
Use the latest 17.2.456
Setup igTree with load on demand with a dataSource with 4 or more nested levels.
### Result  

__depth = ul.attr("data-depth")_ for the second level is evaluated to "11" instead of 2. This behavior goes on for the later levels as well.
### Expected result  
Each node level should have it's corresponding _attr("data-depth")_ set to 0,1,2,3,4 and so on.

### Attachments  
data-depth

IvayloG:
Closes #1564

### DIFF
--- a/src/js/modules/infragistics.ui.tree.js
+++ b/src/js/modules/infragistics.ui.tree.js
@@ -3093,7 +3093,7 @@
 				originalData = this.nodeDataFor(path),
 
 				// I.G. January 31st, 2018 #1564 igTree does not load nodes after lv3 when using loadOnDemand
-				depth = parseInt(ul.attr("data-depth")),
+				depth = parseInt(ul.attr("data-depth"), 10),
 				binding = this._retrieveCurrentDepthBinding(depth - 1),
 				newData = data.data();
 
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.tree.js
```diff
@@ -3091,7 +3091,9 @@
 			var ul = this._populatingNode.ul, node = this._populatingNode.node,
 				path = node.attr("data-path"),
 				originalData = this.nodeDataFor(path),
-				depth = ul.attr("data-depth"),
+
+				// I.G. January 31st, 2018 #1564 igTree does not load nodes after lv3 when using loadOnDemand
+				depth = parseInt(ul.attr("data-depth")),
```
kdinev: This shouldn't pass jshint. You need to provide the radix to `parseInt(x, 10)`.

PR #1603, opened 2018-02-16
#1566 VideoPlayer destroy removes the tooltip correctly

### LINKED ISSUE #1566: Elements still remain in the DOM after the videoplayer is removed
### Description  
Elements still remain in the DOM after the videoplayer is removed

### Steps to reproduce  

1. Run the attached sample;
2. Click on the button to remove the videoplayer.

### Result  
The div with id "player10_ctrls_s_tooltip" still remains on the page.

### Expected result  
The div with id "player10_ctrls_s_tooltip" should be removed.

### Attachments  
[Sample.zip](https://github.com/IgniteUI/ignite-ui/files/1680963/Sample.zip)

ViktorSlavov:
Closes #1566

### DIFF
--- a/src/js/modules/infragistics.ui.shared.js
+++ b/src/js/modules/infragistics.ui.shared.js
@@ -462,7 +462,7 @@
 			if (this.bookmarks) {
 
 				//V.S. February 16, 2018 #1566 VideoPlayer destroy removes the tooltip correctly.
-				$("#" +  this._id("_tooltip")).igTooltip("destroy").remove();
+				$("#" +  this._id("_tooltip")).remove();
 				this.bookmarks.remove();
 			}
 		},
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.shared.js
```diff
@@ -460,6 +460,9 @@
 
 		clearBookmarks: function () {
 			if (this.bookmarks) {
+
+				//V.S. February 16, 2018 #1566 VideoPlayer destroy removes the tooltip correctly.
+				$("#" +  this._id("_tooltip")).igTooltip("destroy").remove();
```
Lipata: [remove() ](https://stackoverflow.com/questions/15939503/does-jquery-remove-also-destroy-resizables-draggables) calls destroy() method automatically, so you don't need to call destroy manually. Thanks!

PR #1643, opened 2018-03-07
#1358 - igDatePicker regional fallback to global regional

### LINKED ISSUE #1358: igDatePicker does not fallback to different than default regional settings.
### Description  
igDatePicker does not fallback to different than default regional settings.

### Steps to reproduce  

1. Open the attached sample
2. Click to open the date picker dropdown

### Result  
The datepicker dropdown - calendar is not in Japanese. 

### Expected result  
The calendar should be in Japanese. Only the ja regional settings are loaded. The widget should fallback to using those regionals.

Note that the display value mask is correct. It uses Japanese regionals.
Another note - if you set regional: "ja" it is working.

### Attachments  
image

[samplee.zip](https://github.com/IgniteUI/ignite-ui/files/1407862/samplee.zip)

ViktorSlavov:
Closes #1358 

igDatePicker now checks for the global regional settings and fallbacks to them if no regional option is passed.

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -11099,6 +11099,13 @@
 			regional = ($.datepicker && typeof reg === "string") ?
 				$.datepicker.regional[ (reg === "defaults" || reg === "en-US") ? "" : reg ] :
 				null;
+
+			//V.S. March 7th 2018 - #1358 if no regional option is provided and a global regional is set, uses the global one
+			if (typeof $.ig.util.regional === "string") {
+				if ($.ig.util.regional !== reg && (reg === "defaults" || reg === "en-US")) {
+					regional = $.datepicker.regional[ $.ig.util.regional ] || $.datepicker.regional[ "" ];
+				}
+			}
 			if (regional === null && $.datepicker) {
 				for (lastRegional in $.datepicker.regional) { }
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -11099,6 +11099,13 @@
 			regional = ($.datepicker && typeof reg === "string") ?
 				$.datepicker.regional[ (reg === "defaults" || reg === "en-US") ? "" : reg ] :
 				null;
+
```
Lipata: Please, combine this setting of the "regional" option with the one above (if possible). First get the language, either from the option, util or set default and the get regional setting for that language

src/js/modules/infragistics.ui.editors.js
```diff
@@ -11095,10 +11095,19 @@
 			this._attachButtonsEvents("dropdown", dropDownButton);
 		},
 		_dpRegion: function () {
-			var reg = this.options.regional, lastRegional, regional;
-			regional = ($.datepicker && typeof reg === "string") ?
-				$.datepicker.regional[ (reg === "defaults" || reg === "en-US") ? "" : reg ] :
-				null;
+			var reg = this.options.regional, lastRegional, regional = null, regAbbr = "";
```
bazal4o: Please rename regAbbr variable to more meaningful one

PR #1667, opened 2018-03-16
#1666 fix: Don't clear editor text in value update method (prevents text ch…

### LINKED ISSUE #1666: textChanged event is not triggered when allowNullValue option is set to true,  editor loses focus and afterwards text is cleared with "x" button
### Description  
textChanged event is not triggered when the allowNullValue option is set to true,  editor loses focus and afterwards text is cleared with "x" button.

### Steps to reproduce  

1) Run the attached sample.
2) Enter some value in the text editor.
3) Click outside of the editor in order to lose focus.
4) Click "x"(clear) button of the editor.
5) Observe the events fired.

Please note that this was working as expected in version 16.2
I am logging this with High severity because the issue appears in Filtering feature, where we the textChanged event is handled and it is visible in the samples browser.
### Result  
Only the valueChanged event is fired.

### Expected result  
Both textChanged and valueChanged events to be fired

### Attachments  
[251229.zip](https://github.com/IgniteUI/ignite-ui/files/1819856/251229.zip)

damyanpetev:
…ange check)

Closes #1666

Removed branch specific to null value that'd clear the visible input too soon. Since that was hit only in this specific setup, should be safe to say this won't affect other functionality, despite the general location.


kacheshmarova:
This issue is still reproducible for numeric editors.

PR #1678, opened 2018-03-21
Update editors references

Lipata:
Update editors references; add version extend;


src/js/modules/infragistics.ui.editors.js
```diff
@@ -21,7 +22,6 @@
 		// AMD. Register as an anonymous module.
 		define( [
 			"./infragistics.ui.widget",
-			"./infragistics.ui.scroll",
 			"./infragistics.ui.validator"
```
kdinev: Shouldn't a notifier dependency be added here as well?

PR #1680, opened 2018-03-22
Update npm deploy tag

damyanpetev:
When the release is tagged as "*beta*" (e.g. 18.1-beta.1, used as package version) the respective npm package will now also be tagged as `next`.
Note the default is usually `latest`, hence the 
`npm i package@latest`
so this will make beta releases available only through direct version install and:
`npm i ignite-ui@next`


damyanpetev:
Actually this should be against 18.1.. will rebase in a bit.

PR #1691, opened 2018-03-30
fix loader issues with categorychart and shapechart 18.1

gmurray81:
Fixes critical issues with the loader wrt to categorychart and shapechart.


kdinev:
@gmurray81 Should I tag a new RC with this fix?

PR #1731, opened 2018-05-21
Datasource remote operations

### LINKED ISSUE #1601: Integrate igTimePicker into the igGrid
Modify the datasource to support time columns. Update filtering algorithm to support time columns.

bkulov:
Closes #1601 

Encode the type of the column in the url (in case of remote operations). This will allow the server to properly identify the column type. Required for time columns as they use date objects.


src/js/modules/infragistics.datasource.js
```diff
@@ -5269,12 +5285,26 @@
...
+						}
+
+						params.summariesParams[ key ] = methodsStr.slice(0, -1);
 					}
 				}
 			}
 		},
+		_getSchemaField: function (fieldName) {
+			if (this.settings.schema && this.settings.schema.fields.find) {
+				return this.settings.schema.fields.find(function (f) { return f.name === fieldName; });
```
mpavlinov: @bkulov : Array.prototype.find is not available in Internet Explorer. This code will throw an error there. I'm not sure we provide a polyfill for it. Can you check it please.

bkulov: Sure. Will modify it.

mpavlinov: @bkulov Agree. It won't throw and error.

bkulov: Done :)

PR #1732, opened 2018-05-30
Fix for TFS bug 251082 and fix failing tests

tbelcheva:
TFS Bug 251082


kdinev:
I inspected the issue and the sample. I actually believe that this is not a bug and the code is functioning properly. `pendingTransactions` are almost exclusively used inside our code, and adding a transaction for a row every time a cell edit inside the row is performed will only cause us to do additional processing.

mpavlinov:
Closing this PR based on Konstantin comments.

PR #1781, opened 2018-08-20
1724 - igTree - Node no longer executes html-encoded scripts on drag

### LINKED ISSUE #1724: If a node text contains scripts in an html-encoded string, they get executed when a node is dragged.
### Description  
If a node text contains scripts in an html-encoded string, they get executed when a node is dragged over.

### Steps to reproduce  
1) Run the attached sample (CASE-28974T_.zip). 
2) Drag the first node over the third node.

### Result  
alert() function gets executed.

### Expected result  
alert() function should not be executed.

### Attachments  
[sample.zip](https://github.com/IgniteUI/ignite-ui/files/2004864/sample.zip)

ViktorSlavov:
Closes #1724

If an html-encoded script tag is passed as node name, the script was executed on drag/hover because hover helper markup was set with `$.text()` instead of `$.html()`. All markup assignments are now done with `$.html()` so jQuery handles escaping;

### DIFF
--- a/src/js/modules/infragistics.ui.tree.js
+++ b/src/js/modules/infragistics.ui.tree.js
@@ -2091,7 +2091,7 @@
 				helper = dragAndDropSettings.helper === "default" ? function (event) {
 					var target = $(event.target).closest("li[data-role=node]"),
 						markup = $(self.options.dragAndDropSettings.invalidMoveToMarkup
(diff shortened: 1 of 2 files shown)


damyanpetev:
@ViktorSlavov  Alright, the test fails correctly and now you can commit the fix as well 😉

PR #1799, opened 2018-09-14
abbreviate and add spaces to financial chart indicator menu

davidnegley:
Closes tfs 257490

### DIFF
--- a/src/css/structure/modules/infragistics.ui.financialchart.css
+++ b/src/css/structure/modules/infragistics.ui.financialchart.css
@@ -141,6 +141,7 @@
 	position: absolute;
 	top: calc(100% + 1px);
 	box-shadow: 0 19px 38px rgba(0,0,0,0.30), 0 15px 12px rgba(0,0,0,0.22);
+	white-space: nowrap;
 }
 
 .financialChartIndicatorMenuDropDownMenu {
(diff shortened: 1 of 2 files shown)


ig-ealbert:
@davidnegley This will affect more than the Angular chart.  If the indicator names are all changed from enumeration values to the regular names, it would be best if they were all localizable.  That would add a lot of entries to the jQuery i18n files.  @yoshimis Do you have any preference for how these indicator menu items are displayed?  (Ex. with or without spaces, EN only or localizable?)

ig-ealbert:
@kdinev @davidnegley I assume the indicator names were changed from enumerations to the actual names on purpose.  If this is the case, then yes, they should be added to the locale files.  But if they can be changed back to the enumerations, that works even better.

davidnegley:
yes, these strings will need to be localized at some point.  they were not localized in 18.1.

davidnegley:
how would it be better to go back to the unspaced enum names?

ig-ealbert:
We don't localize enumerations since the exact enumeration text like "BollingerBands" needs to be used in the code.  But when it's a description term like "Bollinger Bands", then we localize it.

davidnegley:
this does not persuade me to take a step backwards.  i can proceed with making these strings localizable, though.

ig-ealbert:
Yes, that solution is fine.  Please let me or @yoshimis know when they are ready and where they are.

davidnegley:
Okay, the strings are now localizable an are in the usual spot with the other toolbar strings

kdinev:
@davidnegley There's conflict with the master branch. I just merged another PR, which probably caused the conflict. Can you take a look?

PR #1806, opened 2018-09-25
#1804 Remove redundant edit value cache on keydown

### LINKED ISSUE #1804: igEditors don't fire textChanged event on Safari on MacOS
### Description  
Moved from TFS 256852 - originally regarding the igGrid filtering, which relies on the textChanged event.
Safari-specific due to their event order.

### Steps to reproduce  

Text Editor scenario:
1) Open the attached sample in Safari on MacOS - textEditor_sample.html
2) Enable IME input and start typing in Hiragana, Katakana or Romanji.

### Result  
Open the console and note that the textChanged event was not triggered.

### Expected result  
textChanged should be triggered.

### Attachments  
[textEditor_sample.zip](https://github.com/IgniteUI/ignite-ui/files/2416163/textEditor_sample.zip)

damyanpetev:
Closes #1804

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -2749,7 +2749,6 @@
 						self._editorInput.val() !== self._currentInputTextValue) {
 						self._processTextChanged();
 					}
-					self._currentInputTextValue = self._editorInput.val();
 					self._triggerKeyDown(event);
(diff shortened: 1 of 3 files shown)


damyanpetev:
@MayaKirova Since this is related to the grid filtering, can you verify the fix with that scenario?

PR #1838, opened 2018-11-08
Set null value despite min/max values

### LINKED ISSUE #1834: igNumericEditor: minValue setting takes precedence over allowNullValue=true.
### Description  
minValue setting takes precedence over allowNullValue=true.

### Steps to reproduce  
Run the attached sample.

### Result  
Editor value is 5(minValue).

### Expected result  
Editor value is empty(null).

Lipata:
Closes #1834

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -4674,8 +4674,9 @@
 			}
 		},
 		_getValueBetweenMinMax: function(value) {
+
 			// N.A. 7 November 2018, Bug #1834, Initial value that is null, should not be overwritten by the min/max values.
-			if (value !== this.options.nullValue) {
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -4678,6 +4673,18 @@
 				this.options.maxDecimals = this._getOptionOrRegionalValue("minDecimals");
 			}
 		},
+		_getValueBetweenMinMax: function(value) {
+			// N.A. 7 November 2018, Bug #1834, Initial value that is null, should not be overwritten by the min/max values.
+			if (value !== this.options.nullValue) {
```
damyanpetev: And if `allowNullValue` is not enabled?

Lipata: Thanks, I've fixed that and I've added a test.

PR #1859, opened 2018-12-06
Update combo data source properly with KO new data source length is the same

### LINKED ISSUE #1852: Changing combo data source using knockout observable is not working
### Description  
Changing combo data source, using knockout observable, is not working when new data source length is the same as old one

### Steps to reproduce  

1. Open attached demo
2. Press "Change datasource to 6" button

### Result  
Combo list stays unchanges

### Expected result  
Combo list should be updated with the new value

### Attachments  
[igComboEx.zip](https://github.com/IgniteUI/ignite-ui/files/2632638/igComboEx.zip)

Lipata:
Closes #1852

### DIFF
--- a/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
+++ b/src/js/extensions/infragistics.ui.combo.knockout-extensions.js
@@ -76,18 +76,24 @@
         }
     }
 
-    function identicalDataSources(combo, viewModelDataSource) {
-        var valueKey = combo.igCombo("option", "valueKey"), comboDataSource, index;
+    function isNewDataSource(valueAccessor) {
+        var combo = $(valueAccessor().combo),
+            listLength = combo.igCombo("listItems").length,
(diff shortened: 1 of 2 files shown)


src/js/extensions/infragistics.ui.combo.knockout-extensions.js
```diff
@@ -55,6 +55,43 @@
...
+                        options: options
+                    }
+                }, dataSource[ i ]);
+            }
+        }
+    }
+
+    function identicalDataSources(combo, viewModelDataSource) {
+        var valueKey = combo.igCombo("option", "valueKey"), comboDataSource, index;
+
```
bazal4o: If the method gets reused somewhere else it's better to have check for the length of the both dataSources

Lipata: Done

PR #1869, opened 2018-12-11
igVideoPlayer does not throw errors upon initialization on a video element

### LINKED ISSUE #1827: igVideoPlayer throws an error when you initialize it on a video element
### Description  
igVideoPlayer throws an error upon initialization on a `<video>` tag.

### Steps to reproduce  

1. Run the attached sample in a WebKit browser (Chrome, Safari, etc)
2. Open the dev tools

### Result  
The console displays an error message.

### Expected result  
There should be no error messages.

### Attachments  
capture

[videoplayer.zip](https://github.com/IgniteUI/ignite-ui/files/2510202/videoplayer.zip)

jackofdiamond5:
Closes #1827 

~The error occurs because of the untimely use of the `timeupdate` video event. It is bound to the element upon creation and attempts to use the video's duration property. For `WebKit` browsers, such as `Chrome`, the duration property is `NaN` until the `durationchange` event is thrown. So now we wait in an interval for the video to finish loading and then we call the `timeupdate` handler. Albeit possibly inconsistent and error inducing, a workaround would be to bind the control's events when `durationchange` is triggered.~

__Edit:__ The `poster` attribute not being set is what originally triggers the issue, as it becomes `undefined` and chains the two errors together. That is why we don't need to set any intervals or bind events differently. We only need to specify that no action should be taken when the `poster` attribute is not set upon initialization.

### DIFF
--- a/src/js/modules/infragistics.ui.videoplayer.js
+++ b/src/js/modules/infragistics.ui.videoplayer.js
@@ -2508,6 +2508,8 @@
 			this._removeVideoProperty(video, "autoplay");
 			this._removeVideoProperty(video, "preload");
 			this._removeVideoProperty(video, "loop");
+
+			// B.P. Dec 11th, 2018 Bug #1827 An error is thrown when igVideoPlayer is initialized on a video element
 			if (this._oldPoster !== "") {
 				this._removeVideoProperty(video, "poster");
 			}
--- a/src/js/modules/infragistics.ui.videoplayer.js
+++ b/src/js/modules/infragistics.ui.videoplayer.js
@@ -2412,7 +2412,8 @@
 			var o = this.options,
 				video,
 				css = this.css;
-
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.ui.videoplayer.js
```diff
@@ -2501,7 +2508,9 @@
 			this._removeVideoProperty(video, "autoplay");
 			this._removeVideoProperty(video, "preload");
 			this._removeVideoProperty(video, "loop");
-			this._removeVideoProperty(video, "poster");
+			if (this._oldPoster !== "") {
```
jackofdiamond5: The `poster` attribute of the video element becomes `undefined` if it is originally not set. Which in turn reesults in another error. So we take no actions unless it was explicitly set.

PR #1890, opened 2019-01-22
Fix igTimePicker to accept date values

### LINKED ISSUE #1888: igTimePicker does not accepts date value
### Description  
igTimePicker does not accepts date value, even the time portion of value is in the list of drop down values.

### Steps to reproduce  

1. Create igTimePicker like 
            var time = new Date(1,1,1,1,0,0,0);
            $("#igTimePickerSelector").igTimePicker({
                timeInputFormat: "hh:mm tt",
                value: time
            });

2. Open the page with igTimePicker.

### Result  
igTimePicker visualize empty and with warning message "Pick value for the drop-down list"

### Expected result  
igTimePicker should show 1:00 AM as its value

bkulov:
Closes #1888

Introducing breaking change which have to be added to the changelog.

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -12282,7 +12282,7 @@
 				//Set
 				$(".selector").%%WidgetName%%("option", "isLimitedToListValues", false);
 			```*/
-			isLimitedToListValues: true,
+			isLimitedToListValues: false,
 			/* type="bool" Gets/Sets whether the onscreen keyboard (if available on device) should be shown when the dropdown button is clicked/tapped. This option prevents initial focus or removes it when the drop button is clicked/tapped.
 				Note: The option does not perform device detection so its behavior is always active if enabled.
 				Note: When drop down is opened the only way to close it will be using the drop down button.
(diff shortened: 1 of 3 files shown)


src/js/modules/infragistics.ui.editors.js
```diff
@@ -12745,6 +12747,12 @@
 				this._super(this._parseDateFromMaskedValue(currentVal));
 			}
 		},
+		_valueIndexInList: function (val) { //igTimePicker
+			if (!val && val !== 0) {
+				return -1;
+			}
+			return $.inArray(val.toTimeString(), this.options.timeItems);
```
wnvko: Instead of creating second array, timeItems, you can check like this:
return $.inArray(val.toTimeString(), this.options.listItems.toTimeString());
or create a function to compare time values

bkulov: My implementation is better as it caches the items. `_validateValue` is invoked a lot of times and it doesn't make sense to convert list times to strings each time the value needs to be validated. The cache is updated each time the drop down items are recreated.
`$.inArray(val.toTimeString(), this.options.listItems.toTimeString());` is not going to work as `this.options.listItems.toTimeString()` won't return an array.
Creating a function to compare time portions is an overkill as we only need to compare them for equality and not for less or greater values.

PR #1891, opened 2019-01-23
fix(datasource): Fixing issue 1639

### LINKED ISSUE #1639: Data Source settings are not cleared between new instances of the ig.DataSource.
### Description  
Data Source settings are not cleared between new instances of the ig.DataSource.

### Steps to reproduce  

1. Run the attached sample.
2. Note that ds1 has some filtering settings applied.
3. Note that when a new instance of ig.DataSource is created (ds2) it has the same value for its filtering setttings.
### Result  
Settings applies to one instance are applied to new instances.

### Expected result  
New instances should not be poluted by values set on previous instances. They should always be cleared.

### Attachments 
[dsIssue.zip](https://github.com/IgniteUI/ignite-ui/files/1775984/dsIssue.zip)

kacheshmarova:
Data Source settings are not cleared between new instances of the ig.DataSource

Closes #1639

### DIFF
--- a/src/js/modules/infragistics.util.js
+++ b/src/js/modules/infragistics.util.js
@@ -134,7 +134,7 @@
 								typeof _super[ name ] === "function" && fnTest.test(prop[ name ]);
 					if (isFn) {
 						prototype[ name ] = makeFn(name, prop[ name ]);
-					} else if ( name === 'settings' && typeof  prop[ name ] === "object") {
+					} else if ( name === "settings" && typeof  prop[ name ] === "object") {
 						prototype[ name ] = $.extend({}, prop[ name ]);
 					} else {
 						prototype[ name ] = prop[ name ];
--- a/src/js/modules/infragistics.util.js
+++ b/src/js/modules/infragistics.util.js
@@ -125,6 +125,23 @@
 
 		// The dummy class constructor
 		function Class() {
+			prototype = this;
+			for (name in prop) {
+				if (prop.hasOwnProperty(name)) {
+
+					// Check if we're overwriting an existing function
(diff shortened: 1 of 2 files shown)


src/js/modules/infragistics.util.js
```diff
@@ -125,6 +125,23 @@
 
 		// The dummy class constructor
 		function Class() {
+			prototype = this;
+			for (name in prop) {
+				if (prop.hasOwnProperty(name)) {
+
+					// Check if we're overwriting an existing function
+					var isFn = prototype[ name ] = doSuper && typeof prop[ name ] === "function" &&
```
skrustev: Do you need to assign `prototype[ name ]` here as well, even though you set it later?

src/js/modules/infragistics.util.js
```diff
@@ -125,6 +125,23 @@
...
+								typeof _super[ name ] === "function" && fnTest.test(prop[ name ]);
+					if (isFn) {
+						prototype[ name ] = makeFn(name, prop[ name ]);
+					} else if ( name === "settings" && typeof  prop[ name ] === "object") {
+						prototype[ name ] = $.extend({}, prop[ name ]);
```
ChronosSF: I don't like that `settings` becomes a magic property that will get extended while others will not. For example, there is an object property called `schema` in `DataSchema`. Have you tried initializing two `DataSchema`-s and see if the prototype doesn't get polluted?

One way for resolving such issues is already implemented in igGrid and it's basically culling the object props  that are defined for documentation purposes on init (features and columns for the igGrid case). While it's not a clean solution either, it's at least a more targeted fix that will likely not cause butterfly effect issues.

mpavlinov:
@kacheshmarova We decided not to fix this issue.

PR #1914, opened 2019-04-02
1902: igDateEditor - Spinning date w/o previous value present preserves entered value - 19.1

### LINKED ISSUE #1902: TimePicker and DateEditor values reset to 12:00 AM when key up/down is pressed.
### Description  
TimePicker and DateEditor values reset to 12:00 AM when key up/down is pressed and there is no valid value yet entered in the editor.

### Steps to reproduce  

1. Open http://jsfiddle.net/9fnzy71L/
2. Clear value
3. Type in 08:00
4. Use the right arrow key then press the down arrow to select PM

### Result  
The value is reset to 12:00 AM

### Expected result  
The value should be 08:00 PM

ViktorSlavov:
Closes #1902

### DIFF
--- a/src/js/modules/infragistics.ui.editors.js
+++ b/src/js/modules/infragistics.ui.editors.js
@@ -10357,7 +10357,7 @@
 
 			// N.A. 3/12/2016 Bug #215134: When we enter a value and spin before the date is created, we create a today date, cause everything is empty.
 			if (!this._dateObjectValue && mask.indexOf(unfilled) >= 0) {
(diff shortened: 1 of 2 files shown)


TammyD79:
Is this going in 18.2 also? We need it in that release. The other issue was attached to that version too.

ViktorSlavov:
@TammyD79 it is going to be present in 18.2 as well - here's a link to the [merged PR](https://github.com/IgniteUI/ignite-ui/pull/1915)

PR #1928, opened 2019-05-08
Fix issues with to/from oa date

igandrews:
There were issues with how we were converting to/from OA date when the timezone offset for Dec 30, 1899 differed from the current timezone offset which happens for certain timezones.

### DIFF
--- a/tests/unit/util/util-test.js
+++ b/tests/unit/util/util-test.js
@@ -408,8 +408,15 @@ QUnit.test('[ID19] Test arrayCopy1', function (assert) {
 });
 
 QUnit.test('[ID20] Test OADate', function (assert) {
-	assert.expect(1);
+	assert.expect(7);
 
-	var dt = new Date(2000, 1, 1);
-	assert.equal($.ig.Date.prototype.fromOADate(dt.toOADate()).getTime(), +dt,  "The date should roundtrip");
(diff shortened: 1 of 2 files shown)


tests/unit/util/util-test.js
```diff
@@ -405,4 +405,11 @@ QUnit.test('[ID19] Test arrayCopy1', function (assert) {
...
 	$.ig.util.arrayCopy1(arr, 0, target, 1, 4);
 	assert.deepEqual(target, [6,1,2,3,4], "Copy content to the end of another array");
-});
\ No newline at end of file
+});
+
+QUnit.test('[ID20] Test OADate', function (assert) {
+	assert.expect(1);
+
+	var dt = new Date(2000, 1, 1);
```
kdinev: @igandrews Maybe write a few more asserts with a larger variety of dates.

igandrews: I'll add more tests. How come grunt verify didn't catch the errors with the newline in the test?

kdinev: I think tests are excluded from linting.

igandrews: Shouldn't it verify everything so one doesn't need to rely on waiting for the PR results?

kdinev: It totally should... The thing is, if I turn it on right now, there will be A LOT of lint errors :S

PR #2135, opened 2021-07-07
feat(*): adding use strict to JS files #2075

### LINKED ISSUE #2075: JS codes work "use strict"?
## Description
I needs to run bundled code (infragistics.core.js, infragistics.lob.js, infragistics.excel-bundled.js, etc...)  in strict mode on the browser (Chrome/Edge).

To confirm this behavior, I added "use strict" to the beginning of some code file and executed it, but I often get errors such as references to variables that do not have "var" declarations or objects that do not have properties.

Is it possible to make almost all of the bundle code work "use strict"?

http://cdn-na.infragistics.com/igniteui/2020.1/latest/js/infragistics.core.js
http://cdn-na.infragistics.com/igniteui/2020.1/latest/js/infragistics.lob.js

None of the above files declare "use strict" at the beginning of the file. So no error occurs.
The problem is when I try to read these files as "use strict".
That is, whether or not this occurs when "use strict" is added to the beginning of the file.

## Expected result  
I hope to read in strict mode without causing an error that the bundled JS files, such as infragistics.core.js and infragistics.lob.js.
This means that adding a "use strict" declaration to the beginning of a bundle JS file such as infragistics.core.js or infragistics.lob.js will not cause an error in the browser.

kdinev:
Closes #2075 

### Additional information (check all that apply):
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [x] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code
 - [ ] This PR includes API docs for newly added methods/properties
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR contains breaking changes
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

### DIFF
--- a/src/js/modules/infragistics.util.js
+++ b/src/js/modules/infragistics.util.js
@@ -20,6 +20,7 @@
 // Inspired by base2 and Prototype
 
 /*global xyz, Class, ActiveXObject, Modernizr, VBArray, Intl, XDomainRequest, unescape, $, igRoot*/ /*jshint -W106*/ /*jshint -W116*/ /*jshint unused:false*/
+"use strict";
 (function (factory) {
 	if (typeof define === "function" && define.amd) {
 
@@ -55,7 +56,7 @@
 
 	// The base Class implementation (does nothing) or expects Class to already be defined as a function
 	// K.D. August 18, 2016 Bug #242 global scope Class object is overridden by Ignite UI for jQuery Class object
-	this.Class = this.Class || function () { };
+	window.Class = window.Class || function () { };
 
 	// Create a new Class that inherits from this class
 	Class.extend = function (prop, doAugment) {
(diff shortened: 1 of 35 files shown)


src/js/modules/infragistics.util.js
```diff
@@ -55,7 +56,7 @@
 
 	// The base Class implementation (does nothing) or expects Class to already be defined as a function
 	// K.D. August 18, 2016 Bug #242 global scope Class object is overridden by Ignite UI for jQuery Class object
-	this.Class = this.Class || function () { };
+	window.Class = window.Class || function () { };
```
ChronosSF, line 59: I am pretty sure these two lines are not identical and using window here would cause some issue. What I am not sure about is if it weren't something related to Webpack and/or SystemJS and how we should test it (if we even care) ... I'll need some time to get this in the closed source and produce some test scenarios.

kdinev: Without strict mode `this` in anonymous functions is the global object - in this case `window`. With strict mode, this is `undefined` in an anonymous function.

ChronosSF (approved):

PR #2274, opened 2025-02-11
fix(*): replacing deprecated jQuery calls

ChronosSF:
Closes #ISSUE_NUMBER  

### Additional information (check all that apply):
 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code
 - [ ] This PR includes API docs for newly added methods/properties
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR contains breaking changes
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

### DIFF
--- a/src/js/modules/infragistics.ui.shared.js
+++ b/src/js/modules/infragistics.ui.shared.js
@@ -421,7 +421,7 @@
 					" ui-igslider-vertical" +
 					" ui-igslider-disabled")
 				.removeData("slider")
-				.unbind(".slider");
+				.off(".slider");
 
 			this._mouseDestroy();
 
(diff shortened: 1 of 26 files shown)


src/js/modules/infragistics.ui.shared.js
```diff
@@ -720,7 +720,7 @@
 					this.handles.attr("disabled", "disabled");
 					this.element.addClass("ui-disabled");
 				} else {
-					this.handles.removeAttr("disabled");
+					this.handles[0].removeAttribute("disabled");
```
kdinev, line 723: Why is this needed if `attr()` is still supported?

kdinev (approved):

src/js/modules/infragistics.ui.shared.js
ChronosSF: Apparently `removeAttr` should no longer be used when handling boolean attributes. It's being reported by jQuery migrate.

For tags that understand e.g. `disabled` (such as input), we could use removeProp which is the suggested way but in our code we freely use these on divs and such and removeProp actually sets them to false . (yeah it's a mess)

